PR #5, opened 2023-01-02
feat: add bug_report template

max397574:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,103 @@
...
+                config = true,
+          },
+          -- add any other plugins here
+        }
+        require("lazy").setup(plugins, {
+          root = root .. "/plugins",
+        })
+
+        vim.cmd.colorscheme("tokyonight")
+        -- add anything else here
```
UnrealApex, line 100: I see this is a copy of Folke's issue template. Perhaps remove this section because it is out of context here.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,103 @@
...
+  - type: textarea
+    attributes:
+      label: Expected Behavior
+      description: A concise description of what you expected to happen.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Directory structure
+      description: The structure of the directorie used to reproduce the bug
```
UnrealApex, line 56:
```suggestion
      description: The structure of the directory used to reproduce the bug
```
If this typo is in the Lazy repo perhaps I should fix it.

UnrealApex (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
max397574: this isn't lazy specific
it just installs the plugin with lazy
other minimal config used packer
but imo it works much better to use lazy

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,103 @@
...
+        end
+        vim.opt.runtimepath:prepend(lazypath)
+
+        -- install plugins
+        local plugins = {
+          "folke/tokyonight.nvim",
+          {
+                "stevearc/oil.nvim",
+                config = true,
+          },
```
stevearc, line 92: calling `require("oil").setup()` is required in order to set up all the proper autocmds. Does that need to be done somewhere? It also _must_ be done during startup (can't be lazy-loaded) otherwise directory hijacking (e.g. `nvim path/to/dir/`) won't work.

stevearc (changes requested):
Thanks for putting this together! One question about the setup config

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,103 @@
...
+          })
+        end
+        vim.opt.runtimepath:prepend(lazypath)
+
+        -- install plugins
+        local plugins = {
+          "folke/tokyonight.nvim",
+          {
+                "stevearc/oil.nvim",
+                config = true,
```
UnrealApex, line 91:
```suggestion
                "stevearc/oil.nvim",
                config = function()
                  require('oil').setup()
                end
```

UnrealApex (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
max397574: config = true will call the setup function
lazy.nvim automatically detects the main lua module in a plugin and just calls setup in it when you set config to true

max397574:
I tried to make the whole thing a bit clearer for non lazy.nvim users

stevearc (approved):
Looks good, thanks!

.github/ISSUE_TEMPLATE/bug_report.yml
UnrealApex: Did not know that. Thank you for clarifying!

PR #23, opened 2023-01-08
Change prompt option ordering

UnrealApex:
Perhaps yes should be the first answer?


stevearc:
The ordering was intentional. Discarding the changes is currently irreversible, and performing any destructive operation needs to have some friction in front of it to prevent accidents. I think this is a good default policy, but you can override it yourself if you want to forego that protection. All you have to do is change the `<C-l>` binding:

```lua
require("oil").setup({
  keymaps = {
    ["<C-l>"] = "<CMD>edit!<CR>",
  }
})
```

PR #35, opened 2023-01-13
fix: Error when saving blank lines and quitting.

### LINKED ISSUE #34: bug: Error when saving empty lines and then quitting using :wq

- [X] I have read the docs

### Neovim version (nvim -v)

v0.8.2

### Operating system/version

MacOS 13.1

### Getting this error:
```bash
Error executing vim.schedule lua callback: ...te/pack/packer/start/oil.nvim/lua/oil/mutator/parser.lua:49: attempt to index loc
al 'line' (a nil value)
stack traceback:
        ...te/pack/packer/start/oil.nvim/lua/oil/mutator/parser.lua:49: in function 'parse_line'
        ...re/nvim/site/pack/packer/start/oil.nvim/lua/oil/init.lua:45: in function 'get_cursor_entry'
        ...site/pack/packer/start/oil.nvim/lua/oil/mutator/init.lua:505: in function ''

```

This happens after editing the oil buffer while not adding any files or directories, saving and quitting.
For example if i add an empty new line to the buffer, save and quit the error will appear.

I think I fixed this by adding a nil check for the ```line``` in oil/init.lua, function ```get_entry_on_line```.
I can open a PR for the fix above.

### Steps To Reproduce

1. Open the oil buffer
2. Add a new line
3. Save and exit using :wq.

### Expected Behavior

An error shouldn't appear.

### Directory structure

_No response_

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

MysteryMage:
fixes #34

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -41,6 +41,9 @@ M.get_entry_on_line = function(bufnr, lnum)
   end
 
   local line = vim.api.nvim_buf_get_lines(bufnr, lnum - 1, lnum, true)[1]
+  if not line then
+      return nil
+  end
   local column_defs = columns.get_supported_columns(scheme)
   local parsed_entry, entry = parser.parse_line(adapter, line, column_defs)
   if parsed_entry then


stevearc:
Strange...I can't get the error to trigger. Intuitively I wouldn't expect it to be possible, since if your cursor is on a line in the buffer there _must_ be some text on that line. Nothing wrong with adding a nil-check, though.

MysteryMage:
I think it may be vim.schedule executing the code after neovim closes the buffer and then getting a nil ```line``` value, though i guess it doesn't really matter.

PR #67, opened 2023-03-05
doc: reflects that cursorline can be set via `setup.win_options`

nyngwang:
My intention is that I didn't know that I can set `cursorline` by `setup.win_options`, so making this PR for helping people having the same question in mind. I didn't touch anything in your code.


stevearc:
If you're encountering bugs with the window options, could you file an issue with a detailed repro?

PR #69, opened 2023-03-05
feat: add `setup.view_options.is_always_hidden`

nyngwang:
My intention is that I don't want to see macOS folder view file `.DS_Store` while keeping `show_hidden = true`.

Some points:
- The title cannot be achieved by `is_hidden_file` without `is_excluded` added by this PR. (following the naming convention `is_hidden_file`)
- ~~During my changing the ordering of the conditions in [`M.should_display`](https://github.com/stevearc/oil.nvim/blob/415281090746fcf562ddd15ceaf6f047e50e1981/lua/oil/view.lua#L16), I realized that the better name could be `should_display_hidden`. What do you think?~~ not valid after adding `is_always_hidden`


nyngwang:
@Cottser Well that doesn't make sense to me as `.DS_Store` being a hidden file is a truth. It make more sense to extend the default behaviour as written by the author:

https://github.com/stevearc/oil.nvim/blob/415281090746fcf562ddd15ceaf6f047e50e1981/lua/oil/config.lua#L56-L58

i.e. it's used to define something apparently not hidden as an exception, e.g. `node_module/`. (Yeah, the front-end nightmare)

stevearc:
I'd prefer to make this API a bit more generic. I've added a `set_is_hidden_file` API method that should allow you to change what constitutes a "hidden" file at runtime. You can use this as a replacement for `toggle_hidden` by defining your own action that swaps it from the default implementation to one that returns true if `name == '.DS_Store'`

stevearc:
Right, so for your use case you wouldn't want to use the `toggle_hidden` action at all, and would instead want to switch out the `is_hidden_file` function.

```lua
local hidden_prefix = "."
require("oil").setup({
  keymaps = {
    ["g."] = {
      desc = "Toggle hidden files",
      callback = function()
        hidden_prefix = hidden_prefix == "." and ".DS_Store" or "."
        require("oil").set_is_hidden_file(function(name)
          return vim.startswith(name, hidden_prefix)
        end)
      end,
    },
  },
})
```
This way it toggles back and forth between hiding all files that start with `.` and all files that start with `.DS_Store` (technically you wanted filename == ".DS_Store", but this is close enough in practice and is slightly simpler).

stevearc:
Sure, but could we rename `is_excluded` to `is_always_hidden`?

stevearc:
Thanks!

PR #84, opened 2023-03-29
Feature custom icons

NikolaM-Dev:
# Custom Icons

Possibility to customize the icons for the directories and for the fallback of files not yet supported by nvim web icons.

### DIFF
--- a/lua/oil/columns.lua
+++ b/lua/oil/columns.lua
@@ -205,12 +205,12 @@ if has_devicons then
         end
       end
       if type == "directory" then
-        return { conf.directory, "OilDir" }
+        return { conf.directory or " ", "OilDir" }
       else
         local icon
         local hl
         icon, hl = devicons.get_icon(name)
-        icon = icon or conf.default_file
+        icon = icon or conf.default_file or ""
         return { icon .. " ", hl }
       end
     end,


stevearc (changes requested):
Since this change only affects the icon column, I'd prefer if the configuration could be added there. You could pass it in like so:

```lua
require("oil").setup({
  columns = {
    {"icon", default_file = "", directory = ""}
  }
})
```
and then pull those values off of the `conf` parameter that's [passed in to render()](https://github.com/stevearc/oil.nvim/blob/14a914aef1db4bf4b43d98d1bcf29754ea7997fc/lua/oil/columns.lua#L195)

NikolaM-Dev:
Ready

lua/oil/columns.lua
```diff
@@ -205,12 +205,12 @@ if has_devicons then
...
       end
       if type == "directory" then
-        return { " ", "OilDir" }
+        return { conf.directory, "OilDir" }
       else
         local icon
         local hl
         icon, hl = devicons.get_icon(name)
-        icon = icon or ""
+        icon = icon or conf.default_file
```
stevearc, line 213: These will fail for anyone who specifies "icon" in the config with no options. Can you add a fallback to the original value? You will need to handle the case where conf is nil

lua/oil/config.lua
```diff
@@ -2,7 +2,7 @@ local default_config = {
   -- Id is automatically added at the beginning, and name at the end
   -- See :help oil-columns
   columns = {
-    "icon",
+    { "icon", default_file = "", directory = " " },
```
stevearc, line 5: Since this default config doubles as documentation and these options are more for advanced config, I'd prefer to leave it as "icon" here. The documentation of the new config options can go in https://github.com/stevearc/oil.nvim/blob/14a914aef1db4bf4b43d98d1bcf29754ea7997fc/.github/generate.py#L103

See the lines above for an example of how to use LuaParam. It should be a straightforward change, but if you need to test it locally you can run `python .github/main.py generate`

stevearc (changes requested):
Request for backwards compatibility and shuffling the documentation around

lua/oil/config.lua
NikolaM-Dev: I don't understand what you are talking about...

lua/oil/columns.lua
NikolaM-Dev: Okay

stevearc:
I've taken care of it. Thanks for the PR!

Issue #108, opened 2023-05-21
bug: Can't open directory after using LSP jump-to-definition (Windows)


wrapperup:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

0.9.0

### Operating system/version

Windows 10 22H2 (19045.2965)

### Describe the bug

When using jump to definition with an LSP, neovim likes to use a lowercase drive letter.
eg. `c:\...` instead of `C:\...`

However, when trying to open oil while editing a file in this path, it does not open.

This doesn't happen in WSL, so I guess it's a Windows only bug.

### Steps To Reproduce

Use Windows
1. Open a file with an absolute path, but with a lowercase drive letter. eg: `c:\...` instead of `C:\...`
2. Open oil `require("oil").open()`, observe that it does not open. Syntax highlighting also seems to break.

### Expected Behavior

It should open

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Did you check the bug with a clean config?

- [X] I have confirmed that the bug reproduces with `nvim -u repro.lua` using the repro.lua file above.

PR #109, opened 2023-05-21
fix: opening with lowercase drive letters

wrapperup:
Fixes https://github.com/stevearc/oil.nvim/issues/108


stevearc:
Interesting. Not sure what exactly the root cause is, but there's definitely a lot of room for issues on Windows. I've done some brief testing, but I don't regularly use Windows so I'm not surprised that there are some buggy edge cases. Thanks for the PR!

PR #112, opened 2023-05-24
Respect 'clipboard' in copy_entry_path

davidsierradz:
Using `vim.v.register`, we can respect the user's desire of copying to the unnamed (or unnamedplus) register or default one.

Check `:h v:register`

Thanks!


stevearc:
Thanks for the PR!

PR #129, opened 2023-06-17
(feat) added position customizations

aretrosen:
Currently, most of the customizations for float windows are hard-coded. So I added a few customizations for zindex, row and col position, and anchor. I made this customization as I wanted the oil buffer to open on the right side. So, now I can specify the float option like this, which will open the float window on the right side.
```lua
...
float = {
  padding = 0,
  max_width = 40,
  anchor = "NE",
  row = 0,
  col = function()
    return vim.o.columns
  end,
  zindex = 1200,
}
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ Note that at the moment the ssh adapter does not support Windows machines, and i
 - [toggle_float(dir)](doc/api.md#toggle_floatdir)
 - [open(dir)](doc/api.md#opendir)
 - [close()](doc/api.md#close)
-- [select(opts)](doc/api.md#selectopts)
+- [select(opts, callback)](doc/api.md#selectopts-callback)
 - [save(opts)](doc/api.md#saveopts)
 - [setup(opts)](doc/api.md#setupopts)
 
--- a/doc/api.md
+++ b/doc/api.md
@@ -13,7 +13,7 @@
 - [toggle_float(dir)](#toggle_floatdir)
 - [open(dir)](#opendir)
 - [close()](#close)
-- [select(opts)](#selectopts)
+- [select(opts, callback)](#selectopts-callback)
 - [save(opts)](#saveopts)
 - [setup(opts)](#setupopts)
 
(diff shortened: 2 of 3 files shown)


stevearc (changes requested):
Could you also add some entries to the config 
https://github.com/stevearc/oil.nvim/blob/6f8bf067c09e96d6bff548b5e6addb6d9c25a678/lua/oil/config.lua#L74-L83

like `row = nil` and `col = nil` with comments above them that document that you can pass in constants or functions?

aretrosen:
I added some entries to the config, and I documented what are the options meant to be.

README.md
```diff
@@ -200,6 +200,22 @@ require("oil").setup({
...
+    -- col = <func> means place the floating window at the return value of <func>
+    col = function()
+      return vim.o.columns
+    end,
+    -- anchor can optionally be a string. anchor positions the floating window.
+    -- Can be "NE", "NW", "SE", or "SW". By default, "NW" is used.
+    anchor = "NE",
+    -- z-index can optionally be an integer, by default it is 45.
+    -- Higher the z-index, greater is the stacking order of the floating window.
+    zindex = 80,
```
stevearc, line 218: This part of the readme is actually autogenerated from the config.lua file, so the changes have to be made there instead.

stevearc (changes requested):
Actually, looking at this some more I think that a more flexible solution for the long term would be [this override pattern](https://github.com/stevearc/dressing.nvim/blob/5fb5cce0cbfcedeadbcee43e5674e8c9a9f28d4a/lua/dressing/config.lua#L59-L63) I use in dressing.nvim. It allows for complete customization of all the values at runtime, which makes it easy to change the row/col with the width and anchor to make a specific layout. Another benefit is the [code is pretty simple too](https://github.com/stevearc/dressing.nvim/blob/5fb5cce0cbfcedeadbcee43e5674e8c9a9f28d4a/lua/dressing/input.lua#L269).

README.md
aretrosen: I made the necessary changes in config.lua, instead of README.md. Regarding the override function, I wrote up something similar at first too. The problem I faced was that ```nvim_open_win``` takes up integers or fractions. I deliberately wanted them to take a function as argument, as I wanted the oil window to move right when I resize the editor. I might take a look into nvim-notify, as their floating window seems to move to the right as soon as we resize NeoVim. Do you think that would be a good idea?

aretrosen:
I made the necessary changes in config.lua, instead of README.md. Regarding the override function, I wrote up something similar at first too. The problem I faced was that nvim_open_win takes up integers or fractions. I deliberately wanted them to take a function as argument, as I wanted the oil window to move right when I resize the editor. I might take a look into nvim-notify, as their floating window seems to move to the right as soon as we resize NeoVim. Do you think that would be a good idea?

stevearc (changes requested):
> I deliberately wanted them to take a function as argument, as I wanted the oil window to move right when I resize the editor.

These functions are only called before `nvim_open_win`, so they would only evaluate and set the position/size at the time that the floating window is opened. If you want dynamic resizing, you have to tap into the `VimResized` autocmd and manually call `nvim_win_set_config` to update it. That is definitely out of scope here, but could be done in your own user config.

By doing the row/col configuration using an override function like I linked, you get all of the benefits of your previous approach where the row/col could individually be functions, but with simpler code. I think it guides users in the correct direction, as most people probably don't ever want to set a static value for row/col.

PR #132, opened 2023-06-26
(feat) Added override function for floatwin

aretrosen:
My previous pull request was accidentally closed after I rebased my repository. I have now added an override function for the floating window, allowing us to override the float options as shown below.
```lua
float = {
  padding = 0,
  max_width = 40,
  override = function(conf)
    local opts = {
      col = vim.o.columns - conf.width,
      zindex = 80,
    }
    return vim.tbl_deep_extend("force", conf, opts)
  end,
}
```

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -80,6 +80,19 @@ local default_config = {
     win_options = {
       winblend = 10,
     },
+    -- This is the config that will be passed to nvim_open_win.
+    -- Change values here to customize the layout
+    -- For example, you can use this to position the window on the right
+    -- override = function(conf)
+    --   local opts = {
+    --     col = vim.o.columns - conf.width,
+    --     zindex = 80,
+    --   }
+    --   return vim.tbl_deep_extend("force", conf, opts)
+    -- end,
+    override = function(conf)
+      return conf
+    end,
   },
   -- Configuration for the actions floating preview window
   preview = {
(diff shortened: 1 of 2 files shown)


lua/oil/init.lua
```diff
@@ -273,6 +273,7 @@ M.open_float = function(dir)
     border = config.float.border,
     zindex = 45,
   }
+  win_opts = config.float.override(win_opts)
```
stevearc, line 276: With an API like this it can be difficult for users to notice or remember if the customization is in-place or a return value, so it's nice to support both. If they modify it in place and forget to return the config, we can support that by adding a `or win_opts` to the end of this line.

lua/oil/config.lua
```diff
@@ -80,6 +80,19 @@ local default_config = {
...
+    -- This is the config that will be passed to nvim_open_win.
+    -- Change values here to customize the layout
+    -- For example, you can use this to position the window on the right
+    -- override = function(conf)
+    --   local opts = {
+    --     col = vim.o.columns - conf.width,
+    --     zindex = 80,
+    --   }
+    --   return vim.tbl_deep_extend("force", conf, opts)
+    -- end,
```
stevearc, line 92: I actually think you don't need the example here. This example doesn't really do much visually, and any example that produces a sufficiently differentiated layout is probably going to be long enough that it would better fit in a "recipes" section of the docs.

stevearc (changes requested):
Two small changes requested, then LGTM

aretrosen:
Done!

stevearc:
Thanks!

PR #135, opened 2023-06-29
Use default directory config when opening directories

pseudogram:
I was experiencing a bug where if I open a folder using `vim .`, the directory would still be opened with oil, even though I set default_file_explorer to false.

> I haven't tried recreating this bug with a minimal config yet. 

my config using packer:
```lua
    use {
        'stevearc/oil.nvim',
        config = function()
            require('oil').setup({
                columns = {
                    "icon",
                },
                default_file_explorer = false,
            })
        end
    }
```


stevearc:
LGTM, thanks!

PR #139, opened 2023-07-09
Use standard directory highlight group

Jumbub:
I'm not sure if this was intentional or not, but there is a specific highlight group for directories.
(This highlight group is used by other plugins e.g. netrw, nvim-tree)


stevearc:
I can't remember if this was intentional or not either. Looks good to me!

PR #161, opened 2023-08-24
Use /bin/sh instead of bash.

mahlonsmith:
This allows support of other *nix operating systems where bash is not universally present.

.... I routinely use both FreeBSD and Linux.  In FreeBSD, there is no `/bin/bash` -- if it's installed at all via ports, it's under `/usr/local/bin/bash`.

Switching this to `/bin/sh` seems to work fine with some rudimentary usage between various operating systems.   (On most Linuxes, this is bash anyway.  Or Dash.  Or something bash-like.  On the *BSDs, this is Bourne.  In any event, Oil appears to work across the board so far, and I'm pleased to be able to use it across platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ nvim oil-ssh://[username@]hostname[:port]/[path]
 
 This may look familiar. In fact, this is the same url format that netrw uses.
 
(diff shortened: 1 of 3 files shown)


stevearc:
LGTM thanks!

Issue #169, opened 2023-09-03
feature request: ability to sort the files first


sajjathossain:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

the file tree should list file files first and directories last. it allows easier file access.

stevearc:
I've added an API for configuring the sort order. In your case you will want
```lua
require("oil").setup({
  view_options = {
    sort = {
      { "type", "desc" },
      { "name", "asc" },
    }
  }
})
```

PR #170, opened 2023-09-06
add feature request #169

kayuxx:
Based on request #169, I can see that it provides greater user customizability and is overall good.


stevearc:
Thanks for the PR, but I've addressed the feature request in a more general way with 879d280617045d5a00d7a053e86d51c6c80970be

PR #196, opened 2023-10-13
Escape metacharacters when loading file into cmdline

Pokryton:
Small updates for actions `open_cmd_line` and `open_cmdline_dir`.

- Escape special characters in the file path with `fnameescape()` .
- Invoke `nvim_feedkeys()` with `escape_ks=false` as suggested in `:h nvim_feedkeys()` (otherwise there may be problems when file path contains non-ANSCII character).


stevearc:
Thanks for the PR!

PR #205, opened 2023-10-25
Add win_options.number to readme to show line number column in --float mode

roubaobaozi:
I went hunting through the readme and `:h oil` for how to get line numbers in a --float and couldn't find it, just randomly tried flipping all the win_options and then tried adding `number = true` and it worked!

Just wanted to show it as an option in the readme :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ require("oil").setup({
     list = false,
     conceallevel = 3,
     concealcursor = "nvic",
+    number = false,
   },
   -- Send deleted files to the trash instead of permanently deleting them (:help oil-trash)
   delete_to_trash = false,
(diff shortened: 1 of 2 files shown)


stevearc:
Thanks for the suggestion, but I actually decided to go about this a different way. The reason for the style difference on the floating window is because I was passing `style = "minimal"` to `nvim_open_win`, which sets a lot of options by default. I've removed that in 05a80e24f6c920e29ed741d12ede0840e456a3e4, so the style of all oil windows should align to whatever was set via `win_options` merged with your config defaults.

roubaobaozi:
Oh! Tomatoes tomatoes, sounds great, thank you!

PR #209, opened 2023-10-30
feat: make ".." available in file entries

### LINKED ISSUE #166: question: is there a way to show ".." (parent directory)
I know we can navigate to the parent path using a keybind (default : -), But i prefer having the directory ".." listed in the list. Is there a way to show ".."?

ray-pH:
- will close #166 

achieved this by manually adding ".." to the entries of the listed files

```lua
...
table.insert(entries, { name = "..", type = "directory" });
...
```

by default, it will be grouped with the hidden file, but you can make it always shown by setting the `is_hidden_file` function in the user config
```lua
  view_options = {

    -- ...

    is_hidden_file = function(name, bufnr)
      return (name ~= "..") and vim.startswith(name, ".")
    end,

    -- ...

  },
```

not sure if there are any bad side effects,

### DIFF
--- a/lua/oil/adapters/files.lua
+++ b/lua/oil/adapters/files.lua
@@ -284,6 +284,8 @@ M.list = function(url, column_defs, cb)
           end)
           return
         elseif entries then
+          -- HACK: manually insert ".." to the list of entries
+          table.insert(entries, { name = "..", type = "directory" });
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for the PR, but I chose a different implementation for this.

PR #218, opened 2023-11-07
feat: auto-save files affected by will_rename_files

jtubbenhauer:
When moving or renaming files, my expectation was for all files referencing the changed file to be updated and written in the background. The current implementation seems to open all affected files in buffers, though each buffer needs to be manually written for the job to feel 'complete'.

This proposed change writes all affected buffers after `apply_workspace_edit` has executed. It then deletes all buffers that didn't exist prior to the move/rename action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,9 @@ require("oil").setup({
   -- You can set the delay to false to disable cleanup entirely
   -- Note that the cleanup process only starts when none of the oil buffers are currently displayed
   cleanup_delay_ms = 2000,
+  -- Set to true to autosave buffers that are updated with LSP willRenameFiles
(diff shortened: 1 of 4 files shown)


jtubbenhauer:
> I personally prefer being able to visually confirm the updated files before saving, but this would be a great optional addition! What do you think of adding a config option:
> 
> ```lua
> -- Set to true to autosave buffers that are updated with LSP willRenameFiles
> -- Set to "unmodified" to only save unmodified buffers
> lsp_rename_autosave = false
> ```

That sounds like a good solution. Have been flat out recently but will fix this up ASAP.

stevearc:
Thanks for the PR!

PR #237, opened 2023-11-24
fix: allow bare function `require("oil").open` as argument to `nvim_create_user_command`

heygarrett:
I tried to define a user command like this:

```lua
local oil = require("oil")
vim.api.nvim_create_user_command(
	"Ex",
	oil.open,
	{ desc = "Open file browser" }
)
```

But I got this error:

```
E492: Not an editor command: Oil
Error executing Lua callback: ...garrett/.local/share/nvim/lazy/oil.nvim/lua/oil/util.lua:15: attempt to call method 'match' (a nil value)
stack traceback:
        ...garrett/.local/share/nvim/lazy/oil.nvim/lua/oil/util.lua:15: in function 'parse_url'
        ...garrett/.local/share/nvim/lazy/oil.nvim/lua/oil/init.lua:169: in function 'get_url_for_path'
        ...garrett/.local/share/nvim/lazy/oil.nvim/lua/oil/init.lua:367: in function <...garrett/.local/share/nvim/lazy/oil.nvim/lua/oil/init.lua:363>
```

Wrapping it in an anonymous function resolves the issue:

```lua
local oil = require("oil")
vim.api.nvim_create_user_command(
	"Ex",
	function() oil.open() end,
	{ desc = "Open file browser" }
)
```

The reason seems to be that `open` expects a nil or a string. But when a function is used to define a user command it receives a table of information as an argument.

This PR checks to make sure the argument is a string instead of just checking to make sure it's not nil.

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -165,7 +165,7 @@ M.get_url_for_path = function(dir, use_oil_parent)
   if vim.bo.filetype == "netrw" and not dir then
     dir = vim.b.netrw_curdir
   end
-  if dir then
+  if type(dir) == "string" then
     local scheme = util.parse_url(dir)
     if scheme then
       return dir


stevearc:
There's already an `:Oil` user command that does this...why would you not just use that? Or if you're trying to alias it to `:Ex`, you can reference that command instead of calling into the lua API.

heygarrett:
After years of using netrw `:Ex` is just easiest for me, especially since I'm fickle; I end up trying out a lot of plug-ins, and in every file browser I've tried I keep the command the same, for simplicity.

I could alias `:Ex` to `:Oil`, but I've actually deleted the `:Oil` command in my config. 😅 I try to minimize the number of plug-in commands and keymaps defined by default. I'm big fan of utilizing APIs to define my own commands and keymaps.

heygarrett:
Feel free to close if you don't think it's the right direction. I'm content with wrapping it in an anonymous function. Maybe that's just the more proper way to do it?

Issue #29, opened 2023-01-10
bug: oil.nvim doesn't play well with sessions


heygarrett:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

```
NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@Ventura-arm64.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.8.2/share/nvim"

Run :checkhealth for more info
```

### Operating system/version

macOS 13.1

### Describe the bug

Oil buffers saved in a session are not able to be reopened by Neovim when that session is restored. Instead the following message is displayed (`~/.config/nvim` was the directory open in the oil buffer):

```
"oil:///Users/garrett/.config/nvim/" [Not edited] --No lines in buffer--
```

### Steps To Reproduce

1. `nvim -u repro.lua`
2. `:Oil`
3. `:mksession!`
4. `:q`
5. `nvim -u repro.lua`
6. `:source Session.vim`

### Expected Behavior

The oil buffer that was open before quitting Neovim would reopen.

### Repro

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require"oil".setup() -- add any needed settings here
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

stevearc:
[...]

PR #246, opened 2023-12-04
perf: only execute on current buffer since this event is called on each buffer

### LINKED ISSUE #245: bug: Sometimes opening files when using `mkview` and `loadview` get messed up

- [X] I have read the docs

### Neovim version (nvim -v)

0.9.4 release

### Operating system/version

Arch Linux

### Describe the bug

When using the core `view` features in Neovim for saving folds, cursor position, etc. it seems get get messed up when using Oil. Using a simple example pulled directly from the Neovim docs with `:h loadview` to automatically save and load views when opening C files we can get some bad behavior.

When reaching a certain level of nested directories (4 nested directories) and then opening a file that automatically loads a view, the filetype gets messed up. It seems that when Oil is opening a file it's not simply loading it into a buffer like what would happen if you run `:e <some_file>` so I'm not sure how to further debug or diagnose this problem.

Basically when doing `loadview` on the recommended autocommand events it seems that the filetype is being set to `oil` rather than the normal filetype. If we disable the `loadview` part of the autocommands and just run it manually when the file is loaded with `:loadview`, it seems to work fine. So the view that is being created is not wrong, but the way files are loaded really interfere with this core functionality.

Funny enough this behavior is not always the case. It seems to happen 100% of the time if I open up Oil and then navigate into 4 children directories and open a file.

### Steps To Reproduce

0. `rm -rf .repro`, make sure a completely clean install
1. `mkdir -p nested/nested2/nested3/nested4/test.c`, create directory structure
2. `nvim -u repro.lua`, load neovim with our repro and let the plugins install
3. `:Oil`, open oil and then navigate into our nested folders to get all the way to open `test.c`
4. `:set filetype`, check the filetype and see `filetype=c`. This is because it's the first time so no view was loaded, but when we leave the current view will be saved
5. `:q`, quit the editor
6. `nvim -u repro.lua`, load neovim and navigate back all the way down to `test.c`
7. `:set filetype`, check the filetype and see `filetype=oil`

For further testing you can also go into the `repro.lua` you can follow these steps after the first:

1. Edit `repro.lua` and comment out the line `au BufWinEnter *.c silent! loadview` to disable the automatic loading of views but not remove the automatic saving of views
2. `nvim -u repro.lua`, load neovim and navigate to the `test.c` file
3. `:loadview`, load the view manually after the file is opened
4. `:set filetype`, see that `filetype=c` correctly, so the view that is being saved is not being messed up at all, it seems to be something about how the file is being loaded. seemingly in a non-traditional vim way.

### Expected Behavior

I would expect that the usage of core vim/neovim features such as `mkview` and `loadview` should play nicely with my file browser especially when it's extremely useful when it comes to saving folds, cursor position, local variables, etc.

### Directory structure

```
nested/nested2/nested3/nested4/test.c
```

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  {
    "stevearc/oil.nvim",
    config = function()
      require("oil").setup({
      })
    end,
  },
}
  root = root .. "/plugins",
})

-- set `viewdir` to not conflict with system viewdir
vim.opt.viewdir = vim.env["XDG_STATE_HOME"] .. "/nvim/view/"

-- Automatically save and load views for *.c files
-- Used to save foldes, cursor position, etc.
-- snippet directly from Neovim docs: `:h loadview`
vim.cmd([[
  au BufWinLeave *.c mkview
  au BufWinEnter *.c silent! loadview
]])
```

mehalter:
When saving sessions, the file that you source (i.e. `Session.vim`) executes `doautoall SessionLoadPost` which will execute the autocmd event `SessionLoadPost` on each buffer. So we don't need to manually loop over each buffer ourselves, the autocmd execution actually handles it for us. This causes an exponential growth in the number of times these calculations are made.

Also I found this while investigating #245 but sadly this doesn't fix the problem root there.

EDIT: ~I just pushed another performance improvement to check if `oil` is already loaded by checking the filetype. When restoring a session with `source` like the original issue #29 the filetype is not set to `oil`, but when loading views with `loadview` which also calls `SessionLoadPost` the oil buffers are already loaded and therefore the filetype is already `oil`. This allows us to optimize this out. There might be a better way to check if oil is already loaded in a buffer, if so let me know! Another thing is that with this performance update it circumvents the problem in #245 so that it never happens, but doesn't fix the root issue which seems to be a bug in the `load_oil_buffer` function where even when it's only called on buffer numbers that belong to `oil` buffers, they sometimes clobber details in other buffer numbers which I believe is coming from the fact that those buffers do not actively belong to a window. This should still be investigated by someone more familiar with the codebase. Because this doesn't actually fix the bug in that function, it might not be considered "closing" of #245 even though the symptoms of the problem go away~

EDIT 2: Most up to date information on what this PR does and what it does not do can be found below: https://github.com/stevearc/oil.nvim/pull/246#issuecomment-1841370744

With the most recent changes we can also say this this resolves #245 since it _should_ avoid the issue from ever happening, but it might come up down the road if someone gets into some complicated edge cases when using `mksession`. Chances are this is never going to happen though.

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -1128,6 +1128,9 @@ M.setup = function(opts)
     group = aug,
     pattern = "*",
     callback = function(params)
+      if vim.g.SessionLoad ~= 1 then
+        return
+      end
       local util = require("oil.util")
       local scheme = util.parse_url(params.file)
       if config.adapters[scheme] and vim.api.nvim_buf_line_count(params.buf) == 1 then
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -1129,12 +1129,9 @@ M.setup = function(opts)
     pattern = "*",
     callback = function(params)
       local util = require("oil.util")
-      for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
-        local bufname = vim.api.nvim_buf_get_name(bufnr)
-        local scheme = util.parse_url(bufname)
-        if config.adapters[scheme] and vim.api.nvim_buf_line_count(bufnr) == 1 then
-          load_oil_buffer(bufnr)
-        end
+      local scheme = util.parse_url(params.file)
+      if config.adapters[scheme] and vim.api.nvim_buf_line_count(params.buf) == 1 then
+        load_oil_buffer(params.buf)
       end
     end,
   })


stevearc (changes requested):
This was originally intended for handling session files created with `:mksession`. At the time I didn't know about `:mkview` and `:loadview`, and didn't know that they also fired the same autocmd. What do you think of instead adding an early return at the top of this?

```lua
if vim.g.SessionLoad ~= 1 then
  return 
end
```

Then it should only run when sourcing a session file, not when doing `:loadview` (at least, in my testing the scripts created with `:mkview` don't set this variable). Does that also fix the issue for you?

mehalter:
That change does work and adds another check to avoid the bug described in #245. It's definitely a good approach for avoiding running during `loadview` so I made a commit to add this! I think the other performance improvement commits are still valid as well. Let me know what you think, but I think this PR is ready to go :)

Also just to clarify, this PR does resolve the problem that is described in #245 but does not directly resolve the bug in the `load_oil_buffer` function. If there are oil buffers not currently belonging to a window then it will clobber the current window. This still isn't fixed by this PR but it does a good job at avoiding that from happening.

Is there a chance when saving a session where an oil buffer could be saved while it doesn't belong to a window? If that's possible then this bug could still crop back up in the future if someone is an avid user of `:mksession` potentially.

Mkfox1989 (approved):

mehalter:
I removed my 2nd commit which checks if the filetype is `oil` or not. This is no less safe than checking if the buffer only has a single line. I'm not sure if that check should even be there in the first place. There could be many other reasons that a buffer has a single line that doesn't mean the buffer needs to have oil loaded, but that could be looked into separately.

For now it's best to just keep this PR minimal and implement the fix and performance improvements that make the most sense and are semantically sound. For now this PR only does two things:

1. Makes sure that the `SessionLoadPost` auto command only executes if a full session is loaded and not when `:loadview` is executed
2. Only applies the `SessionLoadPost` auto command on the buffer that the event is called on. When sessions and views are loaded this auto command is executed on each and every buffer, so it is a huge performance hit to loop over each buffer. This performance gain moves execution complexity of the auto command from O(n²) to O(n) where n is the number of buffers getting restored

PR #248, opened 2023-12-06
fix: willRename source path

### LINKED ISSUE #247: bug: LSP willRenameFiles not work

- [X] I have read the docs

### Neovim version (nvim -v)

NVIM v0.9.4/ Build type: Release/ LuaJIT 2.1.1700008891

### Operating system/version

Arch Linux x86_64    Kernel: 6.1.63-1-lts

### Describe the bug

I use Mason to install tsserver  and use lspconfig to setup tsserver (you can see in repro.lua below).

I know since version 2.3.0 you have implemented LSP willRenameFiles to auto change import name if i rename file, but  it doesn't work. if I use [nvim-lsp-file-operations](https://github.com/antosha417/nvim-lsp-file-operations) + nvim-tree, it works.

tolerable (can work around it)

### Steps To Reproduce

1.  Create test folder like this
image
file foo.js
image
file test.js
image
file jsconfig.js
image

2. nvim -u repro.lua test/ , Install typescript-language-server using Mason.
3. use command :LspInfo check if client is attached.
image

4. change foo.js to bar.js and save. In test.js , import still foo
image
image

### Expected Behavior

if I change file name, import will also change.

### Directory structure

test/foo.js
test/test.js
test/jsconfig.js

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
            })
        end,
  },
	-- lsp
	"williamboman/mason.nvim",
	"williamboman/mason-lspconfig.nvim",
	{
		"neovim/nvim-lspconfig",
	},
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here

require("mason").setup()
require("mason-lspconfig").setup {
    ensure_installed = { "tsserver" },
}

local lspconfig = require("lspconfig")
lspconfig.tsserver.setup {}
```

luckasRanarison:
Closes #247.
I think getting the relative path is unnecessary because the glob pattern will always match the full path, and it can mess up with some server filters, for example: tsserver uses `**/*.js` but the relative path omit the parent dir.

### DIFF
--- a/lua/oil/lsp_helpers.lua
+++ b/lua/oil/lsp_helpers.lua
@@ -51,7 +51,7 @@ end
 ---@return nil|{src: string, dest: string}
 local function get_matching_paths(client, path_pairs)
   local filters =
-      vim.tbl_get(client.server_capabilities, "workspace", "fileOperations", "willRename", "filters")
+    vim.tbl_get(client.server_capabilities, "workspace", "fileOperations", "willRename", "filters")
   if not filters then
     return nil
   end
--- a/lua/oil/lsp_helpers.lua
+++ b/lua/oil/lsp_helpers.lua
@@ -13,6 +13,16 @@ local function file_matches(filepath, pattern)
       return false
     end
   end
+
+  if vim.lsp._watchfiles then
+    local glob = pattern.glob
+    local path = filepath
+    if vim.tbl_get(pattern, "options", "ignoreCase") then
+      glob, path = glob:lower(), path:lower()
+    end
+    return vim.lsp._watchfiles._match(glob, path)
+  end
+
   local pat = vim.fn.glob2regpat(pattern.glob)
   if vim.tbl_get(pattern, "options", "ignoreCase") then
     pat = "\\c" .. pat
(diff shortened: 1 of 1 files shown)


stevearc (changes requested):
This may fix the immediate problem, but it will introduce more issues. Any pattern that _doesn't_ start with `**` will fail to match because we're passing in the absolute path to the file. The root issue here is vim's `glob2regpat` does not handle `**` the way we expect. See:
`:echo glob2regpat('*.js')` -> `\.js$`
`:echo glob2regpat('**/*.js')` -> `/.*\.js$`
`:echo glob2regpat('**/**/*.js')` -> `/.*/.*\.js$`

This is technically correct by some definitions, but does not behave the way LSP expects.

I think the best solution we have available right now is to use the private method `vim.lsp._watchfiles._match` when it's available, and fall back to `glob2regpat` when it's not.

https://github.com/neovim/neovim/blob/ca7f8786a0eb578895400e23cd21e25cc0f91800/runtime/lua/vim/lsp/_watchfiles.lua#L83-L89

stevearc:
Ah, I was overly broad. Yes it's true that a pattern of `*.js` also works, but if the pattern doesn't start with a `*` or `**`, then it will fail. Example:

```lua
local pat = vim.fn.glob2regpat("src/*.js")
local match = vim.fn.match("/home/user/project/src/file.js", pat)
print(match) -- -1
```

luckasRanarison:
Oh, I see. But I think it's unlikely to happen, I think `FileOperartionPattern` is meant to be matched with the absolute path. tsserver, rust_analyzer and php actor which support `willRenameFiles` all use `**/*.filetype`. I'll update this later and use `vim.ls._watchfiles` and fallback with both absolute path and relative path as a workaround if you're concerned about this.

stevearc:
Thanks for the PR!

PR #249, opened 2023-12-06
Feature: Implement actions for sending oil entries to quickfix #223

### LINKED ISSUE #223: feature request: sending items to quickfix

### Describe the feature

It would be very useful to provide an action in "oil.actions" that allows sending all the files from the currently opened oil buffer to the quickfix list.

### Provide background

As an example, you want to edit all the files adjacent to the currently opened file. You could open the oil buffer and send all files to quickfix, then simply run a command over all the elements in the quickfix, or just iterate over them.

At the moment, I am using a custom function for this, since I find it much better, in the scenario described above, than opening Telescope or something similar and writing a prompt that matches the desired files. Having this built-in would be very convenient, since you could just add the following keymap to config:
```lua
keymaps = {
  ['<C-q>'] = 'actions.send_files_to_quickfix'
}
```

### Additional details

If this is something that would not go against the goals of the project, and is useful enough to be a part of the bulitin actions, I can implement this myself, since it is quite simple and does not seem like it would require any heavy maintenance in the future.

lpoto:
Closes #223 .

Exposes actions:
 - `actions.send_to_qflist`: Sends files from current oil dir to quickfix and replaces old entries
 - `actions.add_to_qflist`: Sends files from current oil dir to quickfix and keeps old entries
 - `actions.send_to_loclist`: Sends files from current oil dir to location list and replaces old entries
 - `actions.add_to_loclist`: Sends files from current oil dir to location list and keeps old entries
 
 The quickfix actions trigger `QuickFixCmdPre` and `QuickFixCmdPost` autocommands.
 
 The actions do not open the quickfix window, since you may create an autocmd to open it automatically on `QuickFixCmdPost`.

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -337,6 +337,55 @@ M.toggle_trash = {
   end,
 }
 
+M.send_to_qflist = {
+  desc = "Sends files in the current oil directory to the quickfix list, replacing the previous entries.",
+  callback = function()
+    util.send_to_quickfix({
+      directories = false,
+      files = true,
+      target = "qf",
+      mode = " ",
+    })
+  end,
+}
+
+M.add_to_qflist = {
+  desc = "Adds files in the current oil directory to the quickfix list, keeping the previous entries.",
+  callback = function()
+    util.send_to_quickfix({
+      directories = false,
+      files = true,
+      target = "qf",
+      mode = "a",
+    })
+  end,
+}
+
+M.send_to_loclist = {
+  desc = "Sends files in the current oil directory to the location list, replacing the previous entries.",
+  callback = function()
+    util.send_to_quickfix({
+      directories = false,
+      files = true,
+      target = "loclist",
+      mode = " ",
+    })
+  end,
+}
+
+M.add_to_loclist = {
+  desc = "Adds files in the current oil directory to the location list, keeping the previous entries.",
+  callback = function()
+    util.send_to_quickfix({
+      directories = false,
+      files = true,
+      target = "loclist",
+      mode = "a",
+    })
+  end,
+}
+
+
 ---List actions for documentation generation
 ---@private
 M._get_actions = function()
--- a/lua/oil/util.lua
+++ b/lua/oil/util.lua
@@ -697,4 +697,54 @@ M.adapter_list_all = function(adapter, url, opts, callback)
   end)
 end
 
+M.send_to_quickfix = function(opts)
+  if type(opts) ~= "table" then
+    opts = {}
+  end
+  local oil = require("oil")
+  local dir = oil.get_current_dir()
+  if type(dir) ~= "string" then
+    return
+  end
+
+  local qf_entries = {}
+  for i = 1, vim.fn.line("$") do
+    local entry = oil.get_entry_on_line(0, i)
+    if
+      entry
+      and (
+        entry.type == "file" and opts.files ~= false
+        or entry.type == "directory" and opts.directories ~= false
+      )
+    then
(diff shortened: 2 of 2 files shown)


lua/oil/util.lua
```diff
@@ -697,4 +697,54 @@ M.adapter_list_all = function(adapter, url, opts, callback)
   end)
 end
 
+M.send_to_quickfix = function(opts)
```
stevearc, line 700: Could you add a type annotation for `opts`? It helps document what the available options are, and provides some type safety. You may also want to include some default values, or make the type non-nilable.

lua/oil/util.lua
```diff
@@ -697,4 +697,54 @@ M.adapter_list_all = function(adapter, url, opts, callback)
...
+  elseif opts.files ~= false and opts.directories == false then
+    qf_title = "oil files"
+  end
+  if opts.target == "loclist" then
+    vim.fn.setloclist(0, qf_entries, opts.mode)
+    vim.fn.setloclist(0, {}, "a", { title = qf_title })
+  else
+    vim.fn.setqflist(qf_entries, opts.mode)
+    vim.fn.setqflist({}, "a", { title = qf_title })
+  end
```
stevearc, line 746: I think you can achieve this with a single call to setloclist/setqflist

```lua
vim.fn.setqflist({}, opts.mode, { title = qf_title, items = qf_entries })
```

lua/oil/util.lua
```diff
@@ -697,4 +697,54 @@ M.adapter_list_all = function(adapter, url, opts, callback)
...
+    then
+      local qf_entry = {
+        filename = dir .. entry.name,
+        lnum = 1,
+        col = 1,
+        text = entry.name,
+      }
+      table.insert(qf_entries, qf_entry)
+    end
+  end
```
stevearc, line 728: This looks like it adds every entry in the buffer to the quickfix. Maybe we should detect if we're in visual mode and only add the selected entries?

lua/oil/actions.lua
```diff
@@ -337,6 +337,55 @@ M.toggle_trash = {
   end,
 }
 
+M.send_to_qflist = {
+  desc = "Sends files in the current oil directory to the quickfix list, replacing the previous entries.",
+  callback = function()
+    util.send_to_quickfix({
+      directories = false,
+      files = true,
```
stevearc, line 345: I don't think that `directories = true` is going to function correctly with the given logic, but if we don't have an immediate use for these parameters (seems like it's always `directories = false` and `files = true`), then I'd err on the side of removing them.

lua/oil/util.lua
```diff
@@ -697,4 +697,54 @@ M.adapter_list_all = function(adapter, url, opts, callback)
...
+  if #qf_entries == 0 then
+    vim.notify_once("[oil] No entries found to send to quickfix", vim.log.levels.ERROR)
```
stevearc, line 730: I think this should be a regular `vim.notify`

stevearc (changes requested):
Nice work! Left a few comments. Please re-request review when you're ready for me to take another look

lua/oil/util.lua
lpoto: It only sends selected items when in visual mode now. I reused the logic you implemented in `lua/oil/init.lua:459` and gathered it in  `get_visual_range` function in util module.

lua/oil/util.lua
lpoto: It now defaults to replacing items in the quickfix list.

stevearc:
Looks good, thanks!

PR #250, opened 2023-12-07
feat: add 'update_on_cursor_moved' option to preview window

umlx5h:
By default, the preview window is updated when the cursor is moved, but this behavior makes it heavy and sometimes stuck when opening large files.
So I added an option to make it behave like netrw's p key, which does not automatically synchronize.

I thought it would be good to allow users to choose which behavior they prefer.


stevearc:
LGTM thanks!

PR #258, opened 2023-12-19
feat: normal_uneditable configuration

### LINKED ISSUE #257: feature request: Allow permissions columns to be non-editable

### Describe the feature

I'd like to be able to force the `permissions` column to be non editable. This way the first mutable column can be the file name, which means the cursor behaves itself a little better:
- the cursor will always stay in front of the file name, even after navigating to the parent directory
- pressing `0` will put the cursor at the start of the file name instead of at the permissions section

I'm thinking just adding a field to the `permissions` column would be a good way to go.

nice to have

### Additional details

Currently I've just forked the plugin to hard code this behavior. But I'd rather contribute it and make it configurable

benlubas:
closes #257

This is the easiest way that I could make this change. It's not all that pretty, but it works. 

I'm open to ideas about this config variable name. Originally I called it `force_uneditable` but you can edit the value in insert mode still, so I changed it to `normal_uneditable`. That name is not great, but I can't think of a better one at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ require("oil").setup({
     -- "size",
     -- "mtime",
   },
+  -- Prevent these columns from being edited
+  normal_uneditable = { },
   -- Buffer-local options to use for oil buffers
   buf_options = {
     buflisted = false,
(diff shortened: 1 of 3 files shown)


stevearc:
Went about doing this in a slightly different way in 71b1ef5edfcee7c58fe611fdd79bfafcb9fb0531

PR #260, opened 2023-12-21
feat: add `didRename` LSP notification support

### LINKED ISSUE #259: feature request: add LSP handler for `didRenameFiles`

### Describe the feature

```lua
          return {
          }
        end, pairs),
```

### Provide background

strongly desired

### Additional details

_No response_

mehalter:
Similar to the built-in `willRename` LSP client integration, this adds in built-in `didRename` support. This includes a couple refactors of some internal functions as well as making sure that only successful move operations are communicated in the `didRename` notification.

So far I have tested to make sure this works with both `vtsls` and `luals` which provide import/require updates on `didRename` and it works great. I have also verified that the old `willRename` support has not broken by testing with `tsserver`.

Resolves #259 

Here are the LSP spec references used in making this:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didRenameFiles

### DIFF
--- a/lua/oil/lsp_helpers.lua
+++ b/lua/oil/lsp_helpers.lua
@@ -3,6 +3,8 @@ local util = require("oil.util")
 
 local M = {}
 
+---@alias oil.PathPair {src: string, dest: string}
+
 ---@param filepath string
 ---@param pattern lsp.FileOperationPattern
 ---@return boolean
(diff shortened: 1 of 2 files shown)


mehalter:
There are all the rest of the `FileOperations` part of the LSP spec (`willCreateFiles`/`didCreateFiles`/`willDeleteFiles`/`didDeleteFiles` notifications/requests) that servers could implement which would be great to support out of the box. I'll be honest, I have no idea what servers even use these so I'm less inclined that it's even necessary to add. But it would be nice to have that sort of support completely native to oil. For now I think it's enough to just have `willRenameFiles`/`didRenameFiles` support as they are both very commonly used and add more when people ask for them.

I also may find myself bored one day and decide to open another PR to implement them as well 😅

Let me know if you think it's worth opening an issue regarding the rest of the `FileOperations` part of the LSP spec or if we just leave it to this random comment.

PR #261, opened 2023-12-21
Implement all LSP workspace file operations

### LINKED ISSUE #259: feature request: add LSP handler for `didRenameFiles`

mehalter:
This generalizes the current `lsp_helpers.lua` file to support all of the workspace file operations in the LSP spec. This could supercede #260 (since it also closes #259) or it could be merged afterwards.

Let me know what you think of this implementation and if you have any ideas/improvements! I focused mainly on decreasing duplicate code while implementing all of these very similar operations while maintaining "type safety" with lua language server.

I also noticed that you have opened a PR to add this sort of functionality directly into Neovim (https://github.com/neovim/neovim/pull/26516) which is amazing! This could still be useful if we want to use that API when it comes out but support backwards compatibility for some amount of time.

### DIFF
--- a/lua/oil/lsp_helpers.lua
+++ b/lua/oil/lsp_helpers.lua
@@ -3,6 +3,8 @@ local util = require("oil.util")
 
 local M = {}
 
+---@alias oil.PathPair {src: string, dest: string}
+
 ---@param filepath string
 ---@param pattern lsp.FileOperationPattern
 ---@return boolean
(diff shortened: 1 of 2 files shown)


stevearc:
Hey, thanks for putting together this PR! As you noticed, I did attempt to add this logic into Neovim core, but it was decided that it doesn't make sense to merge it right now. They want to wait for something in core that would actually make use of it.

I already had a WIP branch that was making use of that PR, so I just pulled those changes into oil itself and put the whole change up as #264. Let me know if that works for you!

mehalter:
Yeah that sounds great! Super appreciate all of your dedication to your plugins and Neovim 🤗

Hope you have a happy holiday season!

PR #264, opened 2023-12-23
feat: support all LSP workspace file operations

### LINKED ISSUE #259: feature request: add LSP handler for `didRenameFiles`

### LINKED ISSUE #262: question: should willRenameFiles be asynchronous?
I noticed that `willRenameFiles` is being called asynchronously, is this intended? In the LSP spec it's explicit that all of the requests do need to be handled before the files are renamed. In a case where several files are renamed at once and the file manager ends up moving them before the edits are applied we could run into issues. I haven't personally run into this so I'm not sure if it's really a bug yet, but I'm curious if this is being handled in some way. It might be more robust to do these calls synchronously with a timeout rather than asynchronously and introduce a race condition that depends on how fast the language server is

stevearc:
Since https://github.com/neovim/neovim/pull/26516 was not merged, we're pulling the relevant logic into oil.nvim.

Supercedes #260 and #261
closes #259
closes #262


mehalter:
Also correct me if I'm wrong, but I believe this also closes #262 since it moves the client requests to being synchronous

mehalter:
Hey just a heads up, I think there might be something going on with the `didRename` call. It seems to work with `vtsls` but I'm not getting it working with the lua language server. I also double checked the branch I had with all of the operations and it was working with the lua language server.

stevearc:
@mehalter do you have a repro? I just checked it on one of my plugins and it seems to be working. Using LuaLS 3.7.0

mehalter:
Ah I can take a look in a few weeks. Just left for some international travel and won't touch a computer until Jan 9, 2024 😅 If you test it and it seems to work then I may have rushed through the test, but I was pretty sure I was careful. I can happily investigate and open a new issue in the new year if I come back and can't get it working!

stevearc:
No worries, enjoy your travel! Please do file an issue later if you can't get it working

GitMurf:
btw I tested this today and it did not rename my imports in a typescript project. i copied over the updated options for Oil adding in the section for LSP renames and then just renamed a file with Oil as usual and applied it. Was I supposed to do something else special to get the LSP rename to work? thanks a bunch for all the good work!

stevearc:
@GitMurf one thing people often trip over is the LSP has to be started first. If you do a `vim .` and start moving files, the typescript LSP won't be active. You have to first open a typescript file, then move the files around with oil, then it should handle renaming the imports.

If that still doesn't work, file a bug report with a minimal repro config and the layout of a minimal typescript project

GitMurf:
@stevearc thanks for the advice. I definitely had a TS file open but it may have been a different file and not the one I renamed. Do you know if the file you are renaming has to be open? Thanks!

stevearc:
No, one TS file open in the project should be sufficient

mehalter:
@stevearc just got back from my trip and did a test and lua language server is definitely not working. I'm opening a new issue currently

PR #271, opened 2024-01-06
perf(trash_windows): use a single powershell instance for operations

TheLeoP:
As mentioned in #243, trash support on windows is slow for some operations because it spawns a powershell process for every operation (sending 10 files to trash would spawn 10 powershell instances).

This PR starts two powershell intances (one for listing and one for deleting) and communicates with them over stdio (sending 10 files to trash now spawns only 1 powershell instance).

This improves the performance of trash related operations in windows after the first operation (because it's neccesary to start the powershell instance the first time). Maybe it could be preffered to start the powershell instances on startup instead (?).

### DIFF
--- a/lua/oil/adapters/trash/windows.lua
+++ b/lua/oil/adapters/trash/windows.lua
@@ -42,23 +42,84 @@ end
 ---@field Path string
 ---@field OriginalPath string
 
----@param cb fun(err?: string, raw_entries: oil.WindowsRawEntry[]?)
-local get_raw_entries = function(cb)
-  ---@type string?
-  local stdout
+---@type integer?
+local get_entries_jid
+---@type string[]
+local stdout = {}
+local is_reading_data = true
+---@type fun(err?: string, raw_entries: oil.WindowsRawEntry[]?)?
+local get_entries_last_cb
 
+---@param cb fun(err?: string, raw_entries: oil.WindowsRawEntry[]?)
+---@return integer? jid
+local get_entries_start_powershell = function(cb)
   local jid = vim.fn.jobstart({
     "powershell",
     "-NoProfile",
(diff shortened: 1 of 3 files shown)


stevearc (changes requested):
With all the global (module-local) variables, this starts to get difficult to reason about. I'm pretty sure there are race conditions here if one of the methods gets called in quick succession. What I'd like to see is a refactoring to push this new added complexity into a self-contained file that exposes a clean API for this trash adapter to delete and restore files. You can take a look at what I'm doing for SSH connections, because it's very similar. In particular, look at this section
https://github.com/stevearc/oil.nvim/blob/c4cc8240f1c71defcb67c45da96e44b968d29e5f/lua/oil/adapters/ssh/connection.lua#L285-L312

This ensures that the commands get executed sequentially because there's no way for a single process to do two at once.

TheLeoP:
Is this approach ok? I'm not sure about how the initial command are handled (I tried to use coroutines, but got an error because they can't be used across c call boundaries). Also, the API isn't  trash specific, would an trash specific API be preffered?

stevearc (changes requested):
This already looks better! I think for this particular case I'd like to see even more of the powershell trash logic pulled out into a different file. Right now the logic for "check if powershell job is running, create one if not" lives in the windows trash adapter file, and it just makes it a little cluttered. It's already a pretty long file with a decent amount of complexity surrounding trash semantics; it would be nice to put all the complexity of managing long-lived jobs into a different file.

stevearc:
I shuffled around some of the logic. LMK if this makes sense (and if it still works for you)

TheLeoP:
Looks good to me and it's working as expected c:

Issue #236, opened 2023-11-24
feature request: do not close preview when cd into dir


al1-ce:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

Current behaviour:
- press `-`
- press `ctrl+p`
- select dir and press `enter`
- preview closes

Desired behaviour:
- press `-`
- press `ctrl+p`
- select dir and press `enter`
- preview stays open and switches to selected file in cd'd dir

umlx5h:
I found weird behaviour regarding this issue.

1. entering directory when preview window is on
2. pressing `-` to go up, then preview window is still.
3. entering <CR> on `../` to go up, then preview window is gone.

PR #277, opened 2024-01-08
feat: do not close preview when cd into dir

lucaseras:
Implements https://github.com/stevearc/oil.nvim/issues/236

Hi there! I'm a first-timer in this repo, so apologies in advance if I've missed any process for contributing

This setup has been working locally for me, so I decided to push up for some feedback, thanks in advance!

Also happy to add this to config, in case people would rather keep the existing behavior

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -386,10 +386,12 @@ M.open = function(dir)
   end
 
   -- If preview window exists, update its content
-  local preview_win_id = util.get_preview_win()
-  if preview_win_id then
-    M.select({ preview = true })
-  end
+  util.run_after_load(0, function()
+    local preview_win_id = util.get_preview_win()
+    if preview_win_id then
+      M.select({ preview = true })
+    end
+  end)
 end
 
 ---Restore the buffer that was present when oil was opened
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -629,7 +629,6 @@ M.select = function(opts, callback)
 
       if not opts.preview and preview_win and entry_is_file then
         vim.api.nvim_win_close(preview_win, true)
-        vim.api.nvim_set_current_win(prev_win)
       end
 
       if opts.preview then
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -529,6 +529,7 @@ M.select = function(opts, callback)
     vim.api.nvim_win_close(preview_win, true)
   end
   local prev_win = vim.api.nvim_get_current_win()
+  local keep_preview_alive
 
   local scheme, dir = util.parse_url(bufname)
   assert(scheme and dir)
@@ -602,6 +603,14 @@ M.select = function(opts, callback)
         vim.bo[filebufnr].buflisted = true
       end
 
+      -- If a preview window is open, and we are selecting a directory,
+      -- persist the preview window
+      if not opts.preview and preview_win then
+        if vim.endswith(normalized_url, "/") then
+          keep_preview_alive = true
+        end
+      end
+
       local cmd
       if opts.preview and preview_win then
         vim.api.nvim_set_current_win(preview_win)
@@ -642,6 +651,9 @@ M.select = function(opts, callback)
       vim.api.nvim_win_call(prev_win, function()
         M.close()
       end)
+    end
+    if keep_preview_alive then
+      M.select({preview = true})
     end
     finish()
   end)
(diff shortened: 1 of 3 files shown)


lua/oil/init.lua
```diff
@@ -643,6 +652,9 @@ M.select = function(opts, callback)
         M.close()
       end)
     end
+    if keep_preview_alive then
+      M.select({preview = true})
+    end
```
stevearc, line 657: One of the difficulties with this is oil loads asynchronously, so it's entirely possible for us to be in the buffer by this point but the buffer hasn't rendered yet. This is a tricky situation, so I added a utility method `util.run_after_load` that will run a function once the oil buffer is ready for use. Try using that here, with some checks to make sure we're still in the buffer and that there is an entry under the cursor (otherwise we'll display an error in an empty folder). You'll want to mostly copy the logic from here

https://github.com/stevearc/oil.nvim/blob/c4cc8240f1c71defcb67c45da96e44b968d29e5f/lua/oil/view.lua#L411-L422

lua/oil/view.lua
```diff
@@ -426,6 +426,31 @@ M.initialize = function(bufnr)
...
+        local preview_win_id = util.get_preview_win()
+        if preview_win_id then
+          if entry.id ~= vim.w[preview_win_id].oil_entry_id then
+            oil.select({ preview = true })
+          end
+        end
+      end
+    end,
+  })
+
```
stevearc, line 453: If we add the call to `util.run_after_load`, I don't think this autocmd will be necessary anymore

stevearc (changes requested):
Thanks for putting up a PR! Left some comments about a few things to explore

lua/oil/init.lua
```diff
@@ -384,6 +384,12 @@ M.open = function(dir)
   if config.buf_options.buflisted ~= nil then
     vim.api.nvim_buf_set_option(0, "buflisted", config.buf_options.buflisted)
   end
+
+  -- If preview window exists, update its content
+  local preview_win_id = util.get_preview_win()
+  if preview_win_id then
+    M.select({ preview = true })
+  end
```
lucaseras, line 392: One question I'm not sure about is: what to do when `update_on_cursor_moved = false`, specifically when moving into a folder/parent. Some ideas but no strong feelings yet:
1. Update the preview window (what I implemented)
2. Keep preview window open but don't update it
3. Close preview window completely

lucaseras:
This new implementation has a problem where sometimes opening a folder this happens:
image
The entry ids are shown next to the icons. I couldn't quite figure out why this is happening

lua/oil/init.lua
lucaseras: Thank you!

lua/oil/view.lua
lucaseras: Yep, removed it 🎉

lua/oil/init.lua
stevearc: I think either 1 or 2 would be fine. If you use this you'll need to make use of the `run_after_load` utility, though. Right now if you open the preview window and go to the parent directory, it displays a "Could not find entry under cursor" error.

lua/oil/init.lua
```diff
@@ -621,6 +624,12 @@ M.select = function(opts, callback)
         args = { filebufnr },
         mods = mods,
       })
+
+      if not opts.preview and preview_win and entry_is_file then
+        vim.api.nvim_win_close(preview_win, true)
+        vim.api.nvim_set_current_win(prev_win)
```
stevearc, line 630: Is this line necessary? Is it possible that the cursor has changed windows between line 533 and here?

stevearc (changes requested):
Tested it out and it's looking good! Just one small bug to fix and one question

lua/oil/init.lua
lucaseras: Good point, yeah I think we can remove that line — the only case where the current window changes is if `opts.preview`

lucaseras:
Sweet, pushed the changes! Also, is [this](https://github.com/stevearc/oil.nvim/pull/277#issuecomment-1897734138) also happening on your end?

stevearc:
I'm not seeing that issue. It looks to me like the conceallevel isn't being set? Possibly something is interfering with the window options. If you can get a repro LMK and I might be able to look into it.

Thanks for working with me on the PR! I think this is a great addition to oil.

mehalter:
@stevearc @lucaseras I also just tried this out and I do get that bug that was mentioned with the entry ids appearing. It seems to only appear on the current line and doesn't happen for every folder. Still working on reproducing in a minimal environment, but just wanted to give a heads up that something is going on here. It looks like `conceallevel` should be getting set to 3 but is set to 2 in my case (which is what I have it set to in my normal vim options, so maybe it's just not getting set at all to 3 rather than getting overridden).

mehalter:
Ah I can replicate it extremely easily. Opened a new issue! #287

Issue #295, opened 2024-01-28
bug: preview window stays open if you leave Oil buffer with :edit


lucaseras:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

NVIM v0.9.5

### Operating system/version

MacOS 14.3

### Describe the bug

With the preview window open, editing a non-oil file **without** selecting it via Oil (e.g. via `:edit`) does not close the preview window. So you are now editing a non-Oil buffer with the preview window still open

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. nvim -u repro.lua
2. `:Oil`
3. `<C-p>` to open preview window
4. `:e baz.txt` to edit file

### Expected Behavior

Preview window should close since we are not in an oil buffer anymore

### Directory structure

a/foo.txt
baz.txt

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Did you check the bug with a clean config?

- [X] I have confirmed that the bug reproduces with `nvim -u repro.lua` using the repro.lua file above.

PR #296, opened 2024-01-28
fix: close preview window when leaving oil buffer

lucaseras:
Fixes https://github.com/stevearc/oil.nvim/issues/295. Closing the preview window only when selecting an entry is not broad enough to handle when we change the buffer for some other reason (e.g. when doing `:edit`)

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -642,6 +642,7 @@ M.select = function(opts, callback)
       if opts.preview then
         vim.api.nvim_set_option_value("previewwindow", true, { scope = "local", win = 0 })
         vim.w.oil_entry_id = entry.id
+        vim.w.oil_source_win = prev_win
         vim.api.nvim_set_current_win(prev_win)
       end
       open_next_entry(cb)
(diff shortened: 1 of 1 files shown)


lua/oil/init.lua
```diff
@@ -910,6 +906,20 @@ local function load_oil_buffer(bufnr)
...
+
+  if not util.is_oil_bufnr(0) then
+    local preview_win_id = util.get_preview_win()
+    if preview_win_id then
+      local current_win_id = vim.api.nvim_get_current_win()
+      if current_win_id and current_win_id ~= preview_win_id then
+        vim.api.nvim_win_close(preview_win_id, true)
+      end
+    end
+  end
```
lucaseras, line 920: Not sure why the tests are not passing... Using `vim.schedule` here makes them pass, but I'd rather not do that to avoid any flickers

stevearc:
@lucaseras I fiddled with it a bit and tried a slightly different approach. See how that works for you

lua/oil/init.lua
```diff
@@ -908,16 +909,21 @@ end
...
-    local preview_win_id = util.get_preview_win()
-    if preview_win_id then
-      local current_win_id = vim.api.nvim_get_current_win()
-      if current_win_id and current_win_id ~= preview_win_id then
-        vim.api.nvim_win_close(preview_win_id, true)
-      end
+  local oil_source_win = vim.w[preview_win_id].oil_source_win
+  if oil_source_win and vim.api.nvim_win_is_valid(oil_source_win) then
+    local src_buf = vim.api.nvim_win_get_buf(oil_source_win)
+    if vim.bo[src_buf].filetype == "oil" then
```
lucaseras, line 920: could we use `util.is_oil_bufnr` instead of checking `filetype`?

stevearc: sure yeah, I forgot that utility existed

lucaseras:
I like the new approach, thank you! I tested it locally and worked nicely

PR #299, opened 2024-01-31
Add border config for SSH and keymaps help window

iamreinder:
Hi there,

First of all, thanks for Oil.nvim, awesome work!

Call me crazy, but one thing I really like, is to have control over how every window border looks within Neovim. Especially for consistency. I dislike it when one window has rounded borders and another one has double borders, etc.

Oil.nvim provides config options for almost every window, but not all of them. I forked this repo to allow configuring the border of the SSH window and the keymaps help window. If you like, you can merge it into master.

This gives us a total of 5 different borders we can configure, which raises the question if we might need a single configuration for all windows. Maybe a window config with overrides per window type, for those who want fine grained control? Something like this:

```lua
{
    window_style = {
         -- Defaults
         border = "rounded",
         padding = 2,
         -- Overrides
         progress_window = {
              border = "single",
              padding = 4,
         },
         -- etc, etc...
    },
}
```

I'd be willing to work on something like that. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,10 @@ require("oil").setup({
     ["g."] = "actions.toggle_hidden",
     ["g\\"] = "actions.toggle_trash",
   },
+  -- Configuration for the floating keymaps help window
+  keymaps_help = {
+    border = "rounded",
+  },
   -- Set to false to disable all of the above keymaps
   use_default_keymaps = true,
   view_options = {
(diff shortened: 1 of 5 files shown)


stevearc:
I like the idea of a shared config, but these windows all have different purposes and occupy different amounts of space. I think the only values that could be shared are the border and potentially some of the window options. For now, this seems like the simplest way forward

iamreinder:
Seems fair. Thanks for merging 🙂

Issue #117, opened 2023-05-31
feature request: Allow moving file into a new directory when the name contains a slash 


titouancreach:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

I have a file like:

foobar.js

if if rename the file to 

hello/foobar.js

I'd like to create a directory `hello` and put `foobar.js` in it

### Additional details

If it's a feature you may like, i can write a PR

stevearc:
This is a neat idea, but it's more complicated than it first seems. The main issue is the possibility that `hello/foobar.js` or even just `hello/` already exist. It's doable, but I'll have to be real careful about the edge cases.

I'll keep this on the back burner until I get the time to look deeper into it.

rasulomaroff:
+1 for this one!

I've been using `oil.nvim` for quite a while already and absolutely love it! I think this feature will kind of complete all the flow that typically happens when editing/creating files/folders.

NReilingh:
Another fun complication :) -- move file to parent directory by renaming it `../file.txt` or `./../file.txt`. For some reason this (as well as subdirectories) feels very intuitive for me in the context of the rest of oil.nvim's features.

rasulomaroff:
> Another fun complication :) -- move file to parent directory by renaming it `../file.txt` or `./../file.txt`. For some reason this (as well as subdirectories) feels very intuitive for me in the context of the rest of oil.nvim's features.

Absolutely! Or moving it into a directory inside a parent one by `../dir/file.txt`

titouancreach:
another one (for fun): transforming a file to a folder if the file is empty and without extension. Use case: I create a foobar file: hoo noo, i wanter to create a folder but a created an extensionless file instead, rename to foobar/

rasulomaroff:
> if the file is empty and without extension

I think that doesn't really necessary. Directories with dots are perfectly valid. Alternatively, if a file is not empty and you are transforming it into a directory, then this information can easily be shown in the confirmation window.

But I would go with any way @stevearc decides to implement, if he does.

stevearc:
@titouancreach congratulations, your pain is also my pain! And that's a much easier fix, so I've gone ahead and implemented it. Now you can change a file to a dir or vice-versa by adding/removing the trailing `/`. It'll translate that change to a delete + create.

For the other requests in this issue, I'm still looking into it. I have some ideas and a rough plan now, but I just haven't had the unbroken time to sit down and hash it out.

[...]

PR #305, opened 2024-02-06
Added ability to move files into folders (and create if they don't exist)

pi314ever:
This solves feature request #117 without major changes to code.

Added functionality:
- Move an object into a folder
- Creates a new folder if it does not exist before

Known issues:
- Moving a file into an existing directory still shows as a CREATE action. (Purely visual, no data is overwritten.)
- Global path shown in preview is not simplified yet (i.e. `/home/user/oil/data/../test/test.lua` vs `/home/user/oil/test/test.lua`)

Example usage:
`<--` = parent directory of oil buffer
File directory structure:
```
- a/ <--
  - b.x
  - c.y
```
After changing `b.x` to `d1/d2/b.x`:
```
- a/ <--
  - d1/
  | - d2/
  | | - b.x
  - c.y
```
After changing `c.y` to `d1/c.y`:
```
- a/ <--
  - d1/
  | - d2/
  | | - b.x
  | - c.y
```
After changing 'c.y' to `../c.y`:
```
- a/ 
  - d1/ <--
  | - d2/
  | | - b.x
  - c.y
```

EDITS: Added examples and known issues

### DIFF
--- a/lua/oil/mutator/init.lua
+++ b/lua/oil/mutator/init.lua
@@ -76,21 +76,21 @@ M.create_actions_from_diffs = function(all_diffs)
     local parent_url = vim.api.nvim_buf_get_name(bufnr)
     for _, diff in ipairs(diffs) do
       if diff.type == "new" then
-        if diff.id then
-          local by_id = diff_by_id[diff.id]
-          ---HACK: set the destination on this diff for use later
-          ---@diagnostic disable-next-line: inject-field
-          diff.dest = parent_url .. diff.name
-          table.insert(by_id, diff)
-        else
-          -- Parse nested files like foo/bar/baz
-          local pieces = vim.split(diff.name, "/")
-          local url = parent_url:gsub("/$", "")
-          for i, v in ipairs(pieces) do
-            local is_last = i == #pieces
-            local entry_type = is_last and diff.entry_type or "directory"
-            local alternation = v:match("{([^}]+)}")
-            if is_last and alternation then
+        -- Parse nested files like foo/bar/baz
+        local pieces = vim.split(diff.name, "/")
+        local url = parent_url:gsub("/$", "")
+        for i, v in ipairs(pieces) do
+          local is_last = i == #pieces
+          local entry_type = is_last and diff.entry_type or "directory"
+          local alternation = v:match("{([^}]+)}")
+          if is_last then
+            if diff.id then
+              local by_id = diff_by_id[diff.id]
+              ---HACK: set the destination on this diff for use later
+              ---@diagnostic disable-next-line: inject-field
+              diff.dest = parent_url .. diff.name
+              table.insert(by_id, diff)
+            elseif alternation then
               -- Parse alternations like foo.{js,test.js}
               for _, alt in ipairs(vim.split(alternation, ",")) do
                 local alt_url = url .. "/" .. v:gsub("{[^}]+}", alt)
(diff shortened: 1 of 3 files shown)


pi314ever:
I realize I did not address the edge case for when files already exist that was mentioned in the issue. However, there is an argument that this decision should not live in this PR as it applies to renaming things in general and should be accounted for explicitly in the `move` operation. I haven't taken the time to dig into that yet, so I am not sure what the default behavior is.

thiagokokada:
I am testing this branch in my configuration and it is exactly what I needed.

Thanks @pi314ever.

pi314ever:
Hey @stevearc , any idea when you could take a look at this? I would love to have this PR merged to improve the capabilities of oil. Thank you for your time!

stevearc (changes requested):
Sorry for the long wait; since this is a subtle, tricky change that requires a lot of thought I had to wait until I found time when I was both free _and_ not exhausted 😅

As far as the implementation goes, I'm impressed that you were able to find a way to add this functionality with a relatively small diff. This is definitely the simplest way to add the requested feature. I do have some reservations, and to explain I'll have to start with some background.

The big danger with oil, as with any powerful file manager, is data loss. This is why a lot of us get a tiny bit of anxiety when we type `rm -rf`, just enough to make us pause and double check that we're not about to ruin ourselves. With oil, it can be harder to make sure that what you're doing is safe. Sure, you'll see what operations are going to be performed, but it can be tricky to spot potential problems. Consider a couple of these examples:

- You're moving a file out of a directory and deleting the directory
- You're swapping the names of two files
- You're deleting a file and renaming another to replace it

Done in the correct order, everything is fine. Done in the wrong order, data loss. And this is especially tricky because oil is fully declarative, so it's agnostic to the order that you perform actions in the interface. My approach to this so far has been to do everything I can to make it _formally impossible_ to produce a set of mutations that cause _unintentional_ data loss (obviously you can still intentionally delete files). The easiest way to see this in action is if you try to swap the names of two (or more) files. It took me a while to figure out how to handle the general case of that.

The problem that I have with this PR, is it breaks those guarantees and opens the door for unintentional side effects. A simple example:
```
mkdir foo
touch foo/a.txt
touch a.txt
```
Then `nvim .` and edit the buffer to look like this:
```
foo/
```
When you save, it moves `a.txt` to `foo/a.txt`, obliterating the pre-existing file. This is the reason why I said on the original issue that this is so difficult. Right now the only way to reference a file is by putting it into the buffer that represents its parent directory (with an exception for if you're creating files in a new directory, because that doesn't have this problem). Since there's only one way to reference it, we're guaranteed that if any file operations would collide, they would both have to have an entry in that parent buffer. We can easily detect this collision during the buffer parsing and prevent the operations from ever happening. Once we open up the possibility to move directories just by changing the name, it's now possible for _any_ buffer to have conflicts with _any_ other buffer, _including_ buffers that haven't even been loaded yet.

So while this is a useful feature and the implementation is actually quite clean, I can't accept it because it breaks the safety contract. It may be very uncommon to find a situation where this causes a problem, but the difference between a file manager that does the correct thing 100% of the time and one that does the correct thing 99.9% of the time is _huge_.

pi314ever:
Thanks for your response! I truly appreciate the time you took to detail your concerns with this PR. The current method for checking duplicates only works on currently open buffers, and the extension to search in unopened buffers does seem quite difficult with the current implementation of oil. I don't think I will have time to figure out how to work this out myself either, but I hope someone can pick up where I left off. 

Some notes to anyone who wants to tackle this problem: 
- The `check_dupe` function in `lua/oil/mutator/parser.lua` is probably the place to modify for the smallest footprint, but the only implementation I can think of may be a performance bottleneck for large projects and moving large amounts of files. 
- The implementation needs to allow for existence if it is not the ending node of the rename (i.e. `a` and `b` for `a/b/c`), while raising errors for ending nodes if duplicates exist for the ending node. 
- I wrote a preliminary test that should catch the overwriting error case when a successful implementation is created. Feel free to take it for future iterations.

Issue #298, opened 2024-01-30
How to open float when opening neovim on directory


tomasmota:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

Is there a way to set in the configuration for when oil takes over the default file explorer, that it uses the float? or just in general setting float as the default? It would be nice to set a single config and have it apply to both default file explorer and :Oil, without having to add stuff on top.

### What is the significance of this feature?

nice to have

PR #312, opened 2024-02-17
Add option to open in float by default

nihancj:
set to true, Oil will open in float mode when :Oil command is used

Solves #298

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -26,6 +26,8 @@ local default_config = {
     conceallevel = 3,
     concealcursor = "nvic",
   },
+  -- Open Oil in flaot view by default
+  default_floating_win = false,
   -- Send deleted files to the trash instead of permanently deleting them (:help oil-trash)
   delete_to_trash = false,
   -- Skip the confirmation popup for simple operations (:help oil.skip_confirm_for_simple_edits)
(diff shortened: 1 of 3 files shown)


lua/oil/config.lua
```diff
@@ -89,6 +89,7 @@ local default_config = {
   },
   -- Configuration for the floating window in oil.open_float
   float = {
+    open_by_default = false,
```
stevearc, line 92: The `float` table contains values to customize the display of the floating window. I think it would make more sense to put this at the top level and call it something like `default_floating_win`

stevearc (changes requested):
This will handle the case of `:Oil`, but the issue was I think also requesting that it work for the case of `:e .`, which is a bit harder.

nihancj:
Making this work on `:e .`  does seem hard.
Would you be able to give some thoughts on where to start?

Also what about adding a `:Oil --normal` flag that opens oil in normal window mode.

stevearc:
Well this is where we switch the buffer from `path/to/file` to `oil://path/to/file`
https://github.com/stevearc/oil.nvim/blob/29a06fcc906f57894c1bc768219ba590e03d1121/lua/oil/init.lua#L687-L690

You would have to somehow instead restore whatever the previous buffer was in this window (might be difficult or impossible if the buffer had `bufhidden=wipe`), then open the floating oil win.

Instead of `--normal` I think `--no-float` would be better.

PR #314, opened 2024-02-25
feat: add header line showing the current directory

KG1603:
* feat: add a header line showing the current directory. can be configured with with view_options.show_header_line. default is false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,8 @@ require("oil").setup({
   -- Set to false to disable all of the above keymaps
   use_default_keymaps = true,
   view_options = {
+    -- Show header line with the name of the current directory
+    show_header_line = false,
     -- Show files and directories that start with "."
     show_hidden = false,
     -- This function defines what is considered a "hidden" file
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -70,6 +70,8 @@ local default_config = {
   -- Set to false to disable all of the above keymaps
   use_default_keymaps = true,
   view_options = {
+    -- Show header line with the current directory
+    show_header_line = false,
     -- Show files and directories that start with "."
     show_hidden = false,
     -- This function defines what is considered a "hidden" file
--- a/lua/oil/mutator/parser.lua
+++ b/lua/oil/mutator/parser.lua
@@ -166,7 +166,13 @@ M.parse = function(bufnr)
     return diffs, errors
   end
 
-  local lines = vim.api.nvim_buf_get_lines(bufnr, 0, -1, true)
+
+  local idx = 0
+  if config.view_options.show_header_line then
+    -- Start with index 1 instead of 0 to skip header line with the name of the current directory
+    idx = 1
+  end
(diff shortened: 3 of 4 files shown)


stevearc:
The problem I have with this approach is that it puts content into the buffer that is both a different format and cannot be modified. Upon encountering this, people may mistakenly think they can make changes to it. Or they may delete it and cause the first line to get skipped parsing. It just introduces one more interface for unexpected behavior.
I think the proper place for this is the winbar, where it is still at the top of the buffer but is not part of the content.

vimalk78:
using `:oil --float` shows the cwd at the top.

stevearc:
I have added a recipe for this https://github.com/stevearc/oil.nvim/blob/master/doc/recipes.md#show-cwd-in-the-winbar

vimalk78:
is using `Oil --float` mandatory? 
i have to use `vim.keymap.set("n", "-", "<CMD>Oil<CR>", { desc = "Open parent directory" })` without which `-` gives error ('Ex not a command`).  
moreover i have to use `Oil --float` without which the top line does not show path **even with using  [show-cwd-in-the-winbar]( https://github.com/stevearc/oil.nvim/blob/master/doc/recipes.md#show-cwd-in-the-winbar)

so having `winbar` plugin and using above config in oil has no effect. meaning without `Oil -float` the path does not show even with using winbar config.

vimalk78:
personally i prefer not to use `Oil --float`

stevearc:
You're missing some key foundations about how vim and neovim work. I'm not going to debug this for you, but I will give you some of the tools you need.

1. Get very familiar with `:help`. Nearly everything about Neovim is documented, and it's all in the help pages. Want to learn what the winbar is? `:help winbar`
2. Winbar.nvim is a plugin, presumably to make it easier to display things in the winbar. Like a statusline plugin, it's not strictly necessary but can make it easier to get the display you like. Also like a statusline plugin, it will probably conflict with anything else trying to manage the winbar (like the snippet I gave you).
3. To view the value of an option, use `:set`. You can see the winbar with `:set winbar`.
4. To see where an option was last set _from_ use `:verbose set winbar`. If it was set from lua code, you will need to start Neovim as `nvim -V1` in order to get the line and file that set the option.
5. Don't be afraid to look deeper into any of your tools to find out how they work. If you were to search for `win_options`, the [first result](https://github.com/search?q=repo%3Astevearc%2Foil.nvim%20win_options&type=code) will show you how it is used.

vimalk78:
I started using  `winbar.nvim` , i find it useful.
image

but if i disable `winbar.nvim` and use https://github.com/stevearc/oil.nvim/blob/master/doc/recipes.md#show-cwd-in-the-winbar with `Oil --float` i get double
image
and `:verbose set winbar` returns
```
  winbar=%!v:lua.get_oil_winbar()
        Last set from Lua (run Nvim with -V1 for more details)
```

now if i use the recipie and disabled `--float` and i get
image

this is EXACTLY, EXACTLY what i wanted and spent quite some time to get it like this, but i was either not getting it or getting double.

Thank you so much for your help. `Oil` is already my favorite plugin.

PR #317, opened 2024-02-27
feat: add ability to alter lsp file operation timeout

TheNordicMule:
I have a large repo in typescript where 1000ms is not enough time for LSP to resolve my rename file request. Therefore, I want to add a config that allows the user to alter the timeout, while keeping the default of 1000ms

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -37,13 +37,14 @@ local default_config = {
   -- You can set the delay to false to disable cleanup entirely
   -- Note that the cleanup process only starts when none of the oil buffers are currently displayed
   cleanup_delay_ms = 2000,
-  -- Set to true to autosave buffers that are updated with LSP willRenameFiles
-  -- Set to "unmodified" to only save unmodified buffers
-  lsp_rename_autosave = false,
-
-  -- The amount of time LSP waits for file operation before it errors
-  lsp_file_operation_timeout_ms = 1000,
-
+  -- set various LSP file operation options
+  lsp_file_methods = {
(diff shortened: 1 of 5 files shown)


lua/oil/config.lua
```diff
@@ -40,6 +40,10 @@ local default_config = {
   -- Set to true to autosave buffers that are updated with LSP willRenameFiles
   -- Set to "unmodified" to only save unmodified buffers
   lsp_rename_autosave = false,
+
+  -- The amount of time LSP waits for file operation before it errors
+  lsp_file_operation_timeout_ms = 1000,
+
```
stevearc, line 46: Since we're adding another option related to LSP file operations (and we may want more in the future), let's nest this under another table
```lua
lsp_file_methods = {
  timeout_ms = 1000
}
```

We'll also want to move `lsp_rename_autosave` into that table with a backwards compatibility shim and deprecation message, but I can do that after this PR.

Would also recommend rewording the comment to something like "Time to wait for LSP file operations to complete before skipping" to make it clear that we're waiting _on_ the LSP, and that the consequences of a timeout are only that it gets skipped.

stevearc (changes requested):
I'm on board with adding a timeout, just have a few organizational changes requested

lua/oil/config.lua
TheNordicMule: Hi @stevearc, I tried to do all of the above. Let me know if there is anything else I need to change. Thank you!

stevearc:
Thanks for the PR!

PR #318, opened 2024-02-28
feat: run external commands on a file under the cursor

omer-biz:
By doing "g:" on a file a prompt asks us which command to run on the current file under the cursor. This is helpful when we want to run a quick command like `file` or `ldd`.

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -403,8 +403,7 @@ M.run_external = {
       if file == "" then
         return
       end
-      local output = vim.fn.system(command .. file)
-      vim.print(output)
+      vim.api.nvim_command(":! " .. command .. " \"" .. file_path .. "\"")
     end)
   end
 }
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -392,4 +392,21 @@ M._get_actions = function()
   return ret
 end
 
+M.run_external = {
+  desc = "Run an external command on an entry",
+  callback = function()
+    vim.ui.input({ prompt = "command: ", completion = "shellcmd" }, function(command)
+      if command == "" then
+        return
+      end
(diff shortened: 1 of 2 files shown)


omer-biz:
Seems like there is no need to merge this pr the same thing can be accomplished by putting this in your config: 
```lua
vim.keymap.set("n", "g:", function()
  vim.ui.input({ prompt = 'command: ', completion = "shellcmd" }, function(input)
    if input == "" then
      return
    end
    local file_path = oil.get_current_dir() .. oil.get_cursor_entry().name
    if file_path == "" then
      return
    end
    vim.api.nvim_command(":! " .. input .. " \"" .. file_path .. "\"")
  end)
end, { desc = "Run custom commands on the entry under cursor" })
```

rafamadriz:
I was looking for a way to do this and found this PR, this is actually super useful. I vote for 
this to be added upstream. It's true that with a very simple function you can add the functionality, however 
this just plays nicely with the kind of workflow that oil gives you, and I think it would make sense to add it. 

If someone comes across this, add this to also handle when you abort the `ui.input`. In those cases it returns `nil` and it would crash. And wrapped it in a autocmd to only add the keymap in an oil buffer

```lua
vim.api.nvim_create_autocmd({ "FileType" }, {
    pattern = { "oil" },
    callback = function()
        vim.keymap.set("n", "g:", function()
            vim.ui.input({ prompt = "command: ", completion = "shellcmd" }, function(input)
                if input == "" or input == nil then return end
                local file_path = oil.get_current_dir() .. oil.get_cursor_entry().name
                if file_path == "" or file_path == nil then return end
                vim.api.nvim_command(":! " .. input .. ' "' .. file_path .. '"')
            end)
        end, { buffer = 0, desc = "Run custom commands on the entry under cursor" })
    end,
})
```

PR #323, opened 2024-03-04
fix(powershell-connection): make powershell use UTF-8 for stdin and stdout

### LINKED ISSUE #316: [windows - trash] bug: delete can't handle non ascii characters

- [X] I have read the docs

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2397+ga9f578b7a

### Operating system/version

Windows 11

### Describe the bug

When trying to delete a file with non ascii characters on its name with the `delete_to_trash = true` option enabled, oil.nvim willl silently fail.

### Steps To Reproduce

1. `nvim -u repro.lua`
2. `:Oil<cr>`
3. `o` (enter insert mode on a new line)
4. `á<esc>:w<cr>` (create a new file with the name `á`) and `o` (confirm)
5. `dd:w<cr>` (delete the file with the name `á`) and `o` (confirm)

### Expected Behavior

oil.nvim should be able to send the file to the windows trash.

### Directory structure

_No response_

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
	      delete_to_trash = true,
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

TheLeoP:
Fixes #316

### DIFF
--- a/lua/oil/adapters/trash/windows/powershell-connection.lua
+++ b/lua/oil/adapters/trash/windows/powershell-connection.lua
@@ -27,15 +27,12 @@ end
 
 ---@param init_command? string
 function PowershellConnection:_init(init_command)
+  -- 65001 is the UTF-8 codepage
+  -- powershell needs to be launched with the UTF-8 codepage to use it for both stdin and stdout
   local jid = vim.fn.jobstart({
-    "powershell",
-    "-NoProfile",
-    "-NoLogo",
-    "-ExecutionPolicy",
-    "Bypass",
-    "-NoExit",
-    "-Command",
-    "-",
+    "cmd",
+    "/c",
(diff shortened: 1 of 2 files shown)


stevearc:
Thanks for figuring out the fix!

Issue #306, opened 2024-02-08
feature request: Natural sorting


EJIqpEP:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

A possibility to add natural sort so files with numbers are sorted correctly

### Provide background

I have a numbered list of files in the folder. Oil sorts them like this
1
...
10
100
101
...
11

### What is the significance of this feature?

cannot use this plugin without it

### Additional details

I didn't find anything in docs or open issues which allows to sort files. For example nvim-tree has the same problem but you can provide a sorting function to it to implement a custom sorting.

EJIqpEP:
No comment at all? Is this feature implemented/planned? Or the project is abandoned?

lucaseras:
@stevearc we can close this issue! Thanks again for the review

PR #328, opened 2024-03-10
Sort entries with natural sorting

lucaseras:
Does #306. Some screenshots of Oil after this change:

## With default sorting
image

Before, the file `15` would come before `9`. Likewise, `v114.lua` would come before `v24.lua`

## With descending name
image

Happy to make this an optional feature!

### DIFF
--- a/lua/oil/columns.lua
+++ b/lua/oil/columns.lua
@@ -279,6 +279,10 @@ M.register("type", {
   end,
 })
 
+local function pad_number(int)
+  return string.format("%012d", int)
+end
+
 M.register("name", {
   render = function(entry, conf)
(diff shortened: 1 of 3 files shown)


stevearc:
Thanks for the PR! I pulled this locally to profile it and kick the tires. There was a bug in the implementation: if you sort by other columns (e.g. size) the values can be numbers. I moved the natural ordering into the definition of the name column, and then also put it behind a config option because I profiled it as being about an order of magnitude slower. For most situations this is probably fine, but if you work with large directories it could cause pauses, so I want to make sure there's a way to disable it.

LMK if this approach works for you

lucaseras:
Sweet, looks good! Appreciate the help 🙏🏻

PR #340, opened 2024-03-29
fix: Adding in SCP options configuration

ericguin:
This changeset adds in additional SCP options to the config. This allows the user to specify a list of flags to send to the SCP command that will be expanded into each shell command.

The primary driver for this is from newer boxes SSHing into pre 9 openSSH boxes. New openSSH uses sftp server under the hood, rather than the older SCP protocol. If you go into a system that does not have these changes, SCP fails to work. The '-O' command line flag was introduced to resolve this.

Using this change, the user can now pass in `extra_scp_options = {"-O"}` to resolve the issue.


stevearc (changes requested):
I think this is a good idea, but `table.unpack` does not exist in LuaJIT. It instead has the global `unpack` function from lua 5.1. Did you even test this to see if it fixes your issue?

ericguin:
I've been running it for a few weeks without issue, yeah. That's super odd -- I'll go ahead and change it and push.

ericguin:
image
Did a quick test on my luajit -- I appear to have a table.unpack

stevearc:
Possible that your luajit is built with `DLUAJIT_ENABLE_LUA52COMPAT` as specified in http://luajit.org/extensions.html, but I don't think this is guaranteed for all Neovim distributions. I'm kind of surprised that it's not uniform, honestly, but I guess the build system allows for it.

Thanks for the PR!

Issue #326, opened 2024-03-08
bug: [Windows] drive letters are handled incorrectly (entry selection, buffer name)


LunarLambda:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2534+g25124d360

### Operating system/version

Windows 10

### Describe the bug

Presume we're starting from `$USERPROFILE`, i.e. `C:\Users\LunarLambda\`.

`:e oil:///` (or `:e .` and press `-` until one hits the drive letter listing)

I get the following:

```
C/
E/
H/
M/
P/
V/
Y/
```

Selecting an entry, say `C`, doesn't open `C:/`, it opens `$PWD/C:/`, so in my case `C:/Users/LunarLambda/C:/`.

Also, oil drops the `:` from the drive prefix, causing `C:/` to display as `C/`, which is bothersome as it breaks vim functions like `fnamemodify` or `expand` (e.g. `matchstr(expand('%:~:.'), 'oil://\w*\zs.*')` gets me `C/Users/LunarLambda` instead of `~`.)

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

See above

### Expected Behavior

Selecting a drive letter from the listing should navigate to it properly.

I assume the reason this happens is because of Windows' drive-relative-working-directory-paths (i.e. `/path` is like `$PWD/path`).

### Directory structure

N/A

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
[...]

PR #341, opened 2024-03-30
Fixed drive switching on windows

mbwilding:
Proposed fix for the traversal part of issue https://github.com/stevearc/oil.nvim/issues/326

### DIFF
--- a/lua/oil/adapters/files.lua
+++ b/lua/oil/adapters/files.lua
@@ -218,9 +218,9 @@ M.normalize_url = function(url, callback)
     if path == "/" then
       return callback(url)
     else
-      local last_letter = url:match("^oil:///(%a)$")
-      if last_letter then
-        return callback("oil:///" .. last_letter .. "/")
+      local drive = url:match("^oil:///(%a)$")
+      if drive then
+        return callback("oil:///" .. drive .. "/")
       end
     end
   end
--- a/lua/oil/adapters/files.lua
+++ b/lua/oil/adapters/files.lua
@@ -277,7 +277,8 @@ end
 local function list_windows_drives(url, column_defs, cb)
   local fetch_meta = columns.get_metadata_fetcher(M, column_defs)
   local stdout = ""
-  local jid = vim.fn.jobstart({ "wmic", "logicaldisk", "get", "name" }, {
+  -- Hide network drives as remote UNC paths are not supported
+  local jid = vim.fn.jobstart({ "wmic", "logicaldisk", "where", "drivetype!=4", "get", "name" }, {
     stdout_buffered = true,
     on_stdout = function(_, data)
       stdout = table.concat(data, "\n")
--- a/lua/oil/adapters/files.lua
+++ b/lua/oil/adapters/files.lua
@@ -227,7 +227,14 @@ M.normalize_url = function(url, callback)
 
   local os_path = vim.fn.fnamemodify(fs.posix_to_os_path(path), ":p")
   uv.fs_realpath(os_path, function(err, new_os_path)
-    local realpath = new_os_path or os_path
+    local realpath
+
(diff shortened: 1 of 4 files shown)


mbwilding:
I am currently looking into how to fixed mapped drives.

lua/oil/adapters/files.lua
```diff
@@ -213,12 +213,28 @@ end
...
-  if fs.is_windows and path == "/" then
-    return callback(url)
+
+  if fs.is_windows then
+    if path == "/" then
+      return callback(url)
+    else
+      local drive = url:match("^oil:///(%u)$")
+      if drive then
+        return callback("oil:///" .. drive .. "/")
```
mbwilding, line 223: This section handles navigating back to root where the drives are listed and allows navigating into them

lua/oil/adapters/files.lua
```diff
@@ -213,12 +213,28 @@ end
...
   local os_path = vim.fn.fnamemodify(fs.posix_to_os_path(path), ":p")
   uv.fs_realpath(os_path, function(err, new_os_path)
-    local realpath = new_os_path or os_path
+    local realpath
+
+    if fs.is_windows then
+      realpath = os_path or new_os_path
+    else
+      realpath = new_os_path or new_os_path
+    end
```
mbwilding, line 236: This section handles fixed mapped network drives, as it was resolving to the UNC path, in my case the \\ IP address rather than the drive letter that was mapped

mbwilding:
Ok, tested on Windows.
Also tested on Ubuntu to make sure nothing was broken.

lua/oil/init.lua
```diff
@@ -686,7 +686,7 @@ local function maybe_hijack_directory_buffer(bufnr)
   end
   local replaced = util.rename_buffer(
     bufnr,
-    util.addslash(config.adapter_to_scheme.files .. vim.fn.fnamemodify(bufname, ":p"))
+    config.adapter_to_scheme.files .. vim.fn.fnamemodify(bufname, ":p")
```
mbwilding, line 689: Fixes when Windows enters nvim via `nvim .` or a directory, it won't have a double slash. Ubuntu seems to work fine with this change

lua/oil/adapters/files.lua
stevearc: Then if `fs.is_windows` is true we expect `realpath` will only ever be set to `os_path`, because `os_path` is not nil. Can we do this as an early return instead and avoid calling `fs_realpath` entirely?

lua/oil/init.lua
stevearc: Could you go into a bit more detail here? The `addslash` should only add a trailing `/` if none currently exists. Is the problem that on windows the buffer name ends with a `\` so it gets modified to `\/`?

stevearc (changes requested):
Couple questions

mbwilding:
All the tests are passing now.
I put that addslash back in but added a check for `\\` in that function instead.
In regards to your question, it was adding two slashes facing the same way.

lua/oil/init.lua
mbwilding: It was adding two slashes facing the same way.

mbwilding:
I tried the early return, but it needs the logic below that does checks like if it is a directory or not. I just throws buffer errors otherwise.

stevearc:
I'm surprised by the presence of backslashes in certain locations. I'm going to need to pull this locally and poke around a bit, but I don't have a windows machine with me at the moment. Will follow up on this next week.

stevearc:
@mbwilding okay, I think I have a better fix for the double `\\` when opening oil with `nvim .`. LMK what you think

mbwilding:
@stevearc Your changes work as expected, I like it.

PR #353, opened 2024-04-16
Escape all paths for ssh file changes using `vim.fn.shellescape()`

### LINKED ISSUE #352: bug: apostrophe's not supported in unix files over ssh

- [X] I have read the docs

### Neovim version (nvim -v)

NVIM v0.9.5

### Operating system/version

6.6.10-1-MANJARO

### Describe the bug

Attempting to create/rename/move/interact with files or folders with apostrophes in them with `oil-ssh:` causes either a hang, or in the case of a move it seems to just ignore the apostrophe, and throw a corresponding error about missing files.

For instance, when attempting to move `testing'.txt` to `testing' 2.txt` I get the following error:
```
[oil] Error applying actions: 1: mv: missing destination file operand after '/Media/testing.txt /Media/testing 2.txt'
```

### Steps To Reproduce

1. nvim -u repro.lua
2. `G`
3. `dd`
4. `:w`
5. `o`

### Expected Behavior

The file should be moved/renamed/deleted/opened depending on actions

### Directory structure

```
- tmp/
	- testing/
		- 	directory/
			- testing'.txt
```

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
    vim.fn.system({
        "git",
        "clone",
        "--filter=blob:none",
        "--single-branch",
        lazypath,
    })
end

-- install plugins
local plugins = {
    "folke/tokyonight.nvim",
    {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
            })
        end,
    },
    -- add any other plugins here
}
    root = root .. "/plugins",
})

-- add anything else here
vim.cmd([[:!rm /tmp/oil-test/ -r]])
vim.cmd([[:!mkdir -p /tmp/oil-test/]])
vim.cmd([[:!touch "/tmp/oil-test/testing'.txt"]])
vim.cmd([[:e oil-ssh://localhost//tmp/oil-test/]])

local buf = 0
local enter = vim.api.nvim_replace_termcodes("<CR>", true, false, true)
vim.cmd([[:sleep ]])
vim.api.nvim_feedkeys(enter, "n", true)
vim.api.nvim_feedkeys("dd", "n", true)
vim.cmd([[:w]])
```

favilo:
This should fix #352. 

At least it works on my own machines.

### DIFF
--- a/lua/oil/adapters/ssh/sshfs.lua
+++ b/lua/oil/adapters/ssh/sshfs.lua
@@ -61,6 +61,8 @@ local function parse_ls_line(line)
   return name, type, meta
 end
 
+local shellescape = vim.fn.shellescape
+
 ---@param url oil.sshUrl
 ---@return oil.sshFs
 function SSHFS.new(url)
@@ -80,7 +82,7 @@ end
 ---@param callback fun(err: nil|string)
 function SSHFS:chmod(value, path, callback)
   local octal = permissions.mode_to_octal_str(value)
-  self.conn:run(string.format("chmod %s '%s'", octal, path), callback)
+  self.conn:run(string.format("chmod %s %s", octal, shellescape(path)), callback)
 end
 
 function SSHFS:open_terminal()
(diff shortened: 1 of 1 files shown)


stevearc (changes requested):
Agree that we need to escape the paths, but the problem with `shellescape` is that its behavior depends on the _local_ platform. So this won't work at all if your local machine is windows, for example. Helpfully, the ssh adapter only supports remote machines with `/bin/sh`, so we don't have to handle multiple types of escaping.

I think we should use our own escaping logic. I believe a simple find and replace of `'` with `'\''` followed by wrapping the entire string in `''` should do it.

stevearc:
Awesome, thanks for the fix!

PR #365, opened 2024-04-30
Feature: add action to open quickfixlist

kevintraver:
Add action which also opens the quickfix list after sending items to it.

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -347,12 +347,34 @@ M.send_to_qflist = {
   end,
 }
 
+M.send_to_qflist_and_open = {
+  desc = "Sends files in the current oil directory to the quickfix list, replacing the previous entries, and open quickfix window",
+  callback = function()
+    util.send_to_quickfix({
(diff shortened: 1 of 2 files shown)


stevearc:
The proposed addition doesn't scale well. There are already 4 different actions for behavior that could be parameterized (`send_to_qflist`, `add_to_qflist`, `send_to_loclist`, `add_to_loclist`). If we add another boolean `open` then logically we would need to double this number to add a `_and_open` variant to each. Arguably I shouldn't have even split it as much as it currently is, but I definitely don't want to split it out further. The preferred way of doing this would be to define your own keymap that leverages the built-in action. Something like:

```lua
require("oil").setup({
  keymaps = {
    ["<C-q>"] = {
      callback = function()
        require("oil.actions").send_to_qflist.callback()
        vim.cmd.copen()
      end,
      desc = "Send items to quickfix and open the quickfix"
    }
  }
})
```

kevintraver:
That makes sense. Thank you for the explanation and keymap example!

PR #366, opened 2024-05-01
fix: icon column highlight

man0xff:
There is a 'highlight' property in the docs for icon column but it doesn't work. So here is a fix.


stevearc:
Thanks for the PR! Made some tweaks to support `highlight` as a function as well as a string.

PR #391, opened 2024-05-28
feat: add copy filename action

kevintraver:
Adds an action similar to `copy_entry_path`, but instead just copy the filename (without the path).


stevearc:
LGTM thanks!

Issue #390, opened 2024-05-26
bug: E11: Invalid in command-line window; <CR> executes, CTRL-C quits


csponge:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

0.11.0 commit 4f24e1b

### Operating system/version

Fedora 40

### Describe the bug

When opening a file from Oil, and then opening a command window (e.g., `q:`) I receive this error:

Error executing vim.schedule lua callback: .../colten/.local/share/nvim/lazy/oil.nvim/lua/oil/view.lua:219: E11: Invalid in command-line window; <CR> executes, CTRL-C quits

I did some digging, and I think the issue is happening in the **delete_hidden_buffers** function in **lua/oil/view.lua**:
``` lua
M.delete_hidden_buffers = function()
  local visible_buffers, hidden_buffers = get_visible_hidden_buffers()
  if not visible_buffers or not hidden_buffers or not vim.tbl_isempty(visible_buffers) then
    return
  end
  for _, bufnr in ipairs(hidden_buffers) do
    vim.api.nvim_buf_delete(bufnr, { force = true })
  end
  cache.clear_everything()
end
```

Calling `vim.api.nvim_buf_delete` throws the error. Taking a look at the old issues, it looks like it was fixed in a different part of the code base: https://github.com/stevearc/oil.nvim/issues/378. It looks like it was fixed by simply exiting before the call to close. Maybe the same solution could be applied here? 

This is my first time opening an issue, so I would appreciate any feedback you have. Also, I would love to help out in fixing the issue if possible. Thanks for an awesome plugin!

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. nvim -u repro.lua
2. Enter command `:Oil`
3. Open a file
4. Before the cleanup delay expires (2 seconds), enter command `:q`

You should then see the error after the cleanup delay expires.

### Expected Behavior

I wouldn't expect to see anything. Oil should handle the buffers in the background.

### Directory structure

N/A

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
[...]

PR #394, opened 2024-05-31
Return from delete_hidden_buffers when win type is command

csponge:
Fixes an issue where attempting to delete hidden buffers when in a command window throws an error.

Ref: https://github.com/stevearc/oil.nvim/issues/390


stevearc:
Thanks for the PR!

PR #400, opened 2024-06-02
Allow files to be moved to non-existing directories

plneple:
Adds handling for non-existing directories for the MOVE command, similar to how CREATE works

### DIFF
--- a/lua/oil/mutator/init.lua
+++ b/lua/oil/mutator/init.lua
@@ -92,6 +92,20 @@ M.create_actions_from_diffs = function(all_diffs)
     local parent_url = vim.api.nvim_buf_get_name(bufnr)
     for _, diff in ipairs(diffs) do
       if diff.type == "new" then
+        local file_name = diff.name
+        local url = parent_url:gsub("/$", "")
+        --- Parse nested files like foo/bar/baz
+        local dirs = vim.split(diff.name, fs.sep)
+        file_name = table.remove(dirs)
+        for i, v in ipairs(dirs) do
+          url = url .. "/" .. v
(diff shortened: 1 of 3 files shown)


stevearc (changes requested):
This currently has the same problems as #305

```
mkdir foo
touch foo/a.txt
touch a.txt
```
Then nvim . and edit the buffer to look like this:
```
foo/
```
When you save, it moves a.txt to foo/a.txt, obliterating the pre-existing file.

The reason this feature does not exist yet is because of the difficulty in ensuring that these operations are safe and cannot produce unexpected data loss.

plneple:
@stevearc 
I was unaware that this idea has already been implemented, my bad.

There are several potential solutions I was able to come up with:
+ Merge as-is. 
If you replace the last step in a scenario you described with a simle `mv` command the outcome is exactly the same, with no warnings or additional flags.
```
mv a.txt foo/a.txt
```
+ Raise an error when moving a file to an existing directory. This covers how I personally use this feature.
+ Introduce a warning before / or while performing the action, if any file happens to be overwritten. I assume this would require adding not-yet existing functionality to the plugin.

Please let me know if you find any of these acceptable.

stevearc:
The ideal would be that if you `:w` with these changes, the save operation is blocked, an error is displayed, and a diagnostic is added to the offending line. This is currently how it works for similar behavior (e.g. creating two files with the same name). 

Alternatively, there is a subset of this functionality that may be easier to add: allowing these files to be moved into a _new_ directory, but not an existing one. This doesn't produce the same issues with filename collisions, and might be easier to code.

PR #401, opened 2024-06-02
improved `set_sort` documentation

Alex23rodriguez:
There was a mistake in the type: the second `[]` was in the discription.  Also added some comments for clarity


stevearc:
These docs are actually autogenerated from the annotations on the method itself. I made the changes there and regenerated the docs, it should be better now. Thanks for bringing this to my attention!

Issue #303, opened 2024-02-04
feature request: Add an option to allow preview the file content during in oil float window mode


kevintsai:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

An option can toggle the functionality of allowing to preview file content during in oil float window mode, and a preview-in-float window config for this float mode preview.

### Provide background

I use oil often in toggle window mode it is easy to see whether current buffer is under oil file explorer or not, but I also want to see (preview) the file content during the file navigation.
It is useful to helping us to make the next action to the files. (like, which file I should rename? etc.)

### What is the significance of this feature?

strongly desired

stevearc:
This would be possible, we would just have to add some special logic for the floating window mode that resizes the oil float and opens a new floating window for the preview. I would review a PR for this

simonlearnscoding:
This would be awsome as I am also usind it in floating window mode

PR #403, opened 2024-06-03
Feature/preview in floating window

PhilippOesch:
Because I'm a big fan of the floating window, but also of the preview functionality.

This implements #303

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -123,8 +123,8 @@ local default_config = {
     },
     -- gap between oil window and preview window
     preview_gap = 2,
-    -- preview_split: Split direction: "left", "right", "above", "below".
-    preview_split = "left",
+    -- preview_split: Split direction: "auto", "left", "right", "above", "below".
+    preview_split = "auto",
     -- This is the config that will be passed to nvim_open_win.
     -- Change values here to customize the layout
     override = function(conf)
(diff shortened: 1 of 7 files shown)


lua/oil/config.lua
```diff
@@ -121,6 +121,10 @@ local default_config = {
     win_options = {
       winblend = 0,
     },
+    -- gap between oil window and preview window
+    preview_gap = 2,
+    -- preview_split: Split direction: "left", "right", "above", "below".
+    preview_split = 'left',
```
stevearc, line 127: I think the default should be `"auto"`, which choose left/right based on the value of `vim.o.splitright`

stevearc (changes requested):
Visually looks great! Needs stylua applied and some of the tests are no longer passing. I also noticed some bugs if you open oil + preview, then `:tabnew` and open oil + preview again.

Will provide a more thorough code review after these things are addressed

stevearc:
@PhilippOesch I moved some of the logic around and did some other various refactoring. Take a look, kick the tires, and if you're happy with it I'll merge!

PhilippOesch:
> @PhilippOesch I moved some of the logic around and did some other various refactoring. Take a look, kick the tires, and if you're happy with it I'll merge!

The refactoring looks good from my view. Thank you for taking your time and looking at it. If you want, you can merge.

Issue #397, opened 2024-05-31
bug: `~` not respected


ro0gr:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

0.10.0

### Operating system/version

MacOS 14.4.1

### Describe the bug

Pressing the `~` key does not toggle the case of the letter under cursor.

Please see the `~` in the https://neovim.io/doc/user/change.html#simple-change.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. Let's put steps from `repro` section at `/tmp/repro/init.lua`
3. `vim /tmp/repro -u /tmp/repro/init.lua`
5. press `~`

**Actual Result**
Nothing happens

### Expected Behavior

`init.lua` should become `Init.lua`. The first letter should become Uppercase.

### Directory structure

```
init.lua
```

### Repro (untouched)

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Did you check the bug with a clean config?

- [X] I have confirmed that the bug reproduces with `nvim -u repro.lua` using the repro.lua file above.

ro0gr:
Currently I'm not familiar with the code base. I'd appreciate any reference to where I should start investigation from, and hopefuly fix this.

ro0gr:
[...]

PR #406, opened 2024-06-05
feat: notify when changing the current directory

ro0gr:
Right now both `actions.tcd` and `actions.cd` seem to do their job silently. That made me think the `~` key is broken(https://github.com/stevearc/oil.nvim/issues/397).

We can improve UX by providing these actions with info notifications, so the keypresses provide some visual response to the user.

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -102,7 +102,7 @@ local function cd(cmd)
   local dir = oil.get_current_dir()
   if dir then
     vim.cmd({ cmd = cmd, args = { dir } })
-    vim.notify(string.format("Change directory: %s", dir), vim.log.levels.INFO)
+    vim.notify(string.format("CWD: %s", dir), vim.log.levels.INFO)
   else
     vim.notify("Cannot :cd; not in a directory", vim.log.levels.WARN)
   end
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -102,6 +102,7 @@ local function cd(cmd)
   local dir = oil.get_current_dir()
   if dir then
     vim.cmd({ cmd = cmd, args = { dir } })
+    vim.notify(string.format("Change directory: %s", dir), vim.log.levels.INFO)
   else
     vim.notify("Cannot :cd; not in a directory", vim.log.levels.WARN)
   end


ro0gr:
Hm.. Strange... I've just double-checked myself, and executed the `tcd` manually from the command line:
```nvim
:lua vim.cmd({ cmd = "tcd", args = { "/home" } })
```
This way my vim just outputs the notification with the latest directory as expected:
Знімок екрана 2024-06-05 о 23 22 15

However, for some reason when [exactly the same code](https://github.com/stevearc/oil.nvim/blob/bbc0e67eebc15342e73b146a50d9b52e6148161b/lua/oil/actions.lua#L104) executes from the oil codebase, it does not trigger the notification. And it does not even add an entry to the `:messages` 🤔  

I'm not a vim/lua expert, and I still don't know why it behaves this way. But I'm pretty sure to say, that "fix" suggested in this PR is wrong. So closing this for now.

Would appreciate any ideas about why the `actions.tcd` is that silent.

ro0gr:
OK, it seems the difference is the keymap. If I invoke `cd` as a part of a keymap it's being silenced. Seems like a nvim/vim thing.

Since the lack of any feedback for `~` is what I'd like to fix, I'm re-opening this.

stevearc:
I had to think about this a good bit, but I don't think that we want this. The behavior of `:cd` to only print the pwd if the command was typed out seems to be intentional
https://github.com/neovim/neovim/blob/cb6c0fda718e4503fc1bfc49a9fe92411f5f9005/src/nvim/ex_docmd.c#L5842-L5847
I dug through the git history and couldn't find any explicit reasoning; it's always been like that. If I had to guess, I would say it's because typing out `:cd` is treating it like a repl, whereas in a keymap users may not want the message history to be affected, and if they do it's easy enough to add a `:pwd` to the keymap.

For the same reason here, I think that the default action makes sense to keep minimal and not have any impact on message history. If you want to get feedback, you can easily add on to the builtins

```lua
require("oil").setup({
  keymaps = {
    ["`"] = function()
      require("oil.actions").tcd.callback()
      print(vim.fn.getcwd())
    end,
  },
})
```

ro0gr:
@stevearc Thank you for taking this into consideration and for your response!

I agree it does probably make sense to silence the `cd` by default for keymaps. However, in scope of the oil, I think, it has another meaning.

Since the oil is supposed to provide just an "edit like a normal buffer" experience(that I like a lot!), I would expect the `~` key to work as it does for normal buffers. 

However, it does not affect my text anymore, instead it changes my `pwd` and does it silently! This was completely unexpected for me as for the new user, and this looks like a spot of poor UX to me, which I still believe is worth to be improved.

ro0gr:
To express my concern better, let's stop thinking about the `~` for a moment. 

Let's assume any other normal mode key is overwritten by the oil silently. So now the `x` stops removing the char under cursor, and changes the working dir silently. Most likely the user would be confused by thinking the `x` just does not work, but in fact it does, but it does some other job behind the user. 

Even if we ignore the newcomers possible scenario, I can still imagine people pressing the `~` by an accident, w/o realizing they've just changed the directory due to no visual response.

stevearc:
That's a fair point. My main concern about this was that if you make the action notify, then there's no way for the end user to silence it if they don't want that notification, but there _is_ a way for them to add the notification if they _do_ want it. I think I have a way in mind to parameterize actions, so if I can get that into decent shape then that can take care of the problem.

I'll merge this in now and see about parameterizing it later.

ro0gr:
Great! Thank you.

Issue #410, opened 2024-06-11
bug: cannot perform file/directory operations within LSP workspaces


namvnngu:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

v0.10.0

### Operating system/version

MacOS 14.5

### Describe the bug

After updating `oil.nvim` to [the commit](https://github.com/stevearc/oil.nvim/commit/a62ec258d1c99d354b256a2ec39733ad9109d808), I cannot perform file/directory operations within LSP workspaces.

Error shown:

```
E5108: Error executing lua: ...local/share/nvim/lazy/oil.nvim/lua/oil/lsp/workspace.lua:80: attempt to call field 'has' (a nil value)
stack traceback:
        ...local/share/nvim/lazy/oil.nvim/lua/oil/lsp/workspace.lua:80: in function 'get_matching_paths'
        ...local/share/nvim/lazy/oil.nvim/lua/oil/lsp/workspace.lua:273: in function 'will_rename_files'
        .../.local/share/nvim/lazy/oil.nvim/lua/oil/lsp/helpers.lua:84: in function 'will_perform_file_operations'
        ....local/share/nvim/lazy/oil.nvim/lua/oil/mutator/init.lua:393: in function 'process_actions'
        ....local/share/nvim/lazy/oil.nvim/lua/oil/mutator/init.lua:567: in function 'cb'
        ...cal/share/nvim/lazy/oil.nvim/lua/oil/mutator/preview.lua:130: in function 'confirm'
        ...cal/share/nvim/lazy/oil.nvim/lua/oil/mutator/preview.lua:182: in function <...cal/share/nvim/lazy/oil.nvim/lua/oil/mutator/preview.lua:181>
```

### What is the severity of this bug?

breaking (some functionality is broken)

### Steps To Reproduce

1. Open any workspace that has LSP triggered.
2. Open a random file to trigger LSP
3. Try to delete a file

https://github.com/stevearc/oil.nvim/assets/42376176/bde077a0-17ee-4f99-8d1b-c5ca150716b4

### Expected Behavior

I should be able to perform file/directory operations within LSP workspaces. 

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
[...]

PR #411, opened 2024-06-11
fix: cannot perform file/directory operations within LSP workspaces

namvnngu:
Hi @stevearc, 

This is my very first PR to contribute to the most favorite plugin that I cannot live without. I raise a PR to fix the issue #410.
The root cause of the issue is that the plugin called `vim.has("nvim-0.11")` instead of `vim.fn.has("nvim-0.11")` in the file `lua/oil/lsp/workspace.lua`.

I would really appreciate if you can review my PR.

Thank you so much.


RVxLab:
@namvnngu I can confirm your fix worked on 0.10 stable, much appreciated!

namvnngu:
> @namvnngu I can confirm your fix worked on 0.10 stable, much appreciated!

Thanks for your confirmation.

stevearc:
Thanks for the fix!

PR #412, opened 2024-06-11
fix: vim.has doesn't exist, use vim.fn.has instead

faergeek:
It errors when you try to rename a file otherwise


RVxLab:
#411 fixes this as well

stevearc:
Thanks for fixing this, merged #411 which was the same change

PR #414, opened 2024-06-12
fix: incorrect default config actions

k14lb3:
https://github.com/stevearc/oil.nvim/blob/76bfc25520e4edc98d089d023b4ed06013639849/lua/oil/actions.lua#L54-L60

https://github.com/stevearc/oil.nvim/blob/76bfc25520e4edc98d089d023b4ed06013639849/lua/oil/actions.lua#L14-L44


stevearc:
Thanks for the fix!

PR #424, opened 2024-06-18
refactor: move loading message delay to config

kevintraver:
This allows `loading_message_delay` to be overridden

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -46,6 +46,8 @@ local default_config = {
     -- Set to "unmodified" to only save unmodified buffers
     autosave_changes = false,
   },
+  -- Delay the loading screen just a bit to avoid flicker
+  loading_message_delay = 100,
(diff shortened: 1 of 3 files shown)


stevearc:
In general I want to avoid exposing additional options in the config unless they are necessary. The list of options is already quite large, and it makes it hard for users to find what they're looking for. Could you describe the issue you are trying to solve with this, and why this is necessary?

kevintraver:
I sometimes see the loading indicator flash, and would like to set it to a higher value.

Maybe this can be something thats not in config by default, but still in the docs and is can be added manually.

stevearc:
What is a value that would cause it to not flicker for you? 100ms was pulled from nowhere in particular, and I'm _probably_ fine with any number below 500ms. I wouldn't even blink at bumping to 200ms.

kevintraver:
Yeah, it does not flicker at 200ms.

stevearc:
Delay updated to 200ms

PR #428, opened 2024-06-20
feat: improve is_modifiable checks for unix os's

### LINKED ISSUE #342: feature request: set `modifiable` to true if the user is in the directory group

### Describe the feature

Currently, when checking if a buffer should be modifiable, it only checks if the group is the same group as what neovim is opened with, it would be nice if it would check if the user is in the group regardless if they have opened neovim with the specific group that the folder needs.

### Provide background

I'm trying to edit files belonging to another group, but although I'm in the group needed for the directory I'm trying to open it still sets modifiable as false since I haven't opened neovim with the specific group. To get around this I can just set modifiable to true or run `sg GROUPNAME "nvim ...."` but it's quite tedious to remember to do so.

nice to have

### Additional details

Not quite sure how you'd want this to be implemented, we could read from /etc/group and check if the user is part of the group.
Or more simply, split `vim.fn.system("id -G")` and check if the directory's group id is in the result.

I'd be happy to implement this

DerpDays:
## About
This will resolve #342, it improves the `is_modifiable` function for directories on unix systems by more accurately checking whether the user has write permission for the directory.

This PR changes the function so that it:
- Checks the groups that the nvim process owner is a part of, instead of only the group that nvim was launched with, this is done using the `id -G` command (gets all gid's of groups that the user is in) and checking if directory gid is in this list.
    * note: the command output is cached upon first run if successful
    * note: if the command fails it falls back to the previous behavior of only checking if the directory gid equal to the nvim process gid
- Accounts for weirder permission setups e.g. 170 and 117 (where the user is the owner of the directory) by still checking the group and others permission sets, and doing an bitwise OR with all of the permission sets that they match.

### Questions
- Is the way I went about the cache fine? I saw that the current cache implementation was based around paths, however since the command `id -G` is independent of directories (based on the nvim process user), I just added an extra variable to the cache module with functions to get and set it.
- With the current cache, even if the command fails for one directory it will try again for the next directory the user enters, should I just make it so that if it fails the first time to not try again, or can I leave it as is?

### DIFF
--- a/lua/oil/adapters/files.lua
+++ b/lua/oil/adapters/files.lua
@@ -433,14 +433,21 @@ M.is_modifiable = function(bufnr)
   end
 
   local uid = uv.getuid()
-  local gid = uv.getgid()
+  local gids = vim.fn.split(vim.fn.system("id -G"))
   local rwx
   if uid == stat.uid then
     rwx = bit.rshift(stat.mode, 6)
-  elseif gid == stat.gid then
-    rwx = bit.rshift(stat.mode, 3)
   else
-    rwx = stat.mode
+    local in_group = false
+    for _, gid in ipairs(gids) do
+      if tonumber(gid) == stat.gid then
+        rwx = bit.rshift(stat.mode, 3)
+        in_group = true
+      end
+    end
+    if not in_group then
+      rwx = stat.mode
+    end
   end
   return bit.band(rwx, 2) ~= 0
 end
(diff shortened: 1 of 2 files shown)


stevearc:
I refactored the caching logic a bit because we didn't need it to be as fancy. LMK if this looks good to you and I'll merge

DerpDays:
Looks good to me! I had kept it in the cache utility since I wasn't sure on how you'd want it, but this way is a lot easier to follow and nicer!

Issue #422, opened 2024-06-17
feature request: Case insensitive sorting


hertelukas:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

I'm a bit confused that this hasn't been mentioned so far, so maybe I overlooked something.

oil sorts all uppercase files/folders in front of lower case ones. I'm not used to this and would like to change it, as I've never seen it like this before. I noticed that netrw has the same behavior, but after some quick googling, one can set `vim.g.netrw_sort_options = "i"`, to ignore the case. However, this does not seem to affect oil. Is there a similar option?

### What is the significance of this feature?

nice to have

hertelukas:
Thank you guys!

PR #429, opened 2024-06-23
Feature/case insensitive sorting

PhilippOesch:
Add case insensitive name sorting option in config #422


stevearc:
Thanks for the PR!

PR #439, opened 2024-07-05
feat: add support for `mini.icons`

mehalter:
This adds explicit support for [`mini.icons`](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-icons.md) for providing icons to oil. This generalizes the icon providing into a function that figures out which icon provider is available (preferring `mini.icons` over `nvim-web-devicons`) and returning a common icon providing interface.

`mini.icons` also has the ability to provide directory icons so this allows the icon provider to provide any type of icon and it's up to the provider to figure out how to handle what it can.

2024-07-05_09:48:45_screenshot
2024-07-05_09:48:59_screenshot

I tested it with each plugin individually as well as both and neither. It is working great! Let me know what you think of the implementation or if you have any recommendations to refactor this. I tried to match the code style and conventions in the codebase.

Thanks so much for such a great project!

### DIFF
--- a/lua/oil/util.lua
+++ b/lua/oil/util.lua
@@ -8,6 +8,8 @@ local FIELD_NAME = constants.FIELD_NAME
 local FIELD_TYPE = constants.FIELD_TYPE
 local FIELD_META = constants.FIELD_META
 
+---@alias oil.IconProvider fun(type: string, name: string, conf: table?): (icon: string, hl: string)
+
 ---@param url string
 ---@return nil|string
 ---@return nil|string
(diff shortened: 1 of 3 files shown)


stevearc:
Thanks for the PR!

lua/oil/util.lua
```diff
@@ -858,4 +860,30 @@ M.get_edit_path = function(bufnr, entry, callback)
   end
 end
 
+--- Check for an icon provider and return a common icon provider API
+---@return (oil.IconProvider)?
+M.get_icon_provider = function()
+  -- prefer mini.icons
+  local has_mini_icons, mini_icons = pcall(require, "mini.icons")
```
echasnovski, line 867: @mehalter, this is not the recommended way for plugins to determine if 'mini.icons' should be used. See [here](https://github.com/echasnovski/mini.nvim/blob/d187614ecee84ef8d7456339ed7ce1160d054310/doc/mini-icons.txt#L248-L250). The reason is that `pcall(require, 'mini.icons')` would return `true` if user has the whole 'mini.nvim' installed, which would conflict with possible idea of using 'nvim-web-devicons' instead of 'mini.icons'.

mehalter: Ah thanks! I will also check for `_G.MiniIcons` with a follow up. We definitely want to do the `require` at least to protect against the user doing any sort of lazy loading.

mehalter: Hm I suppose if the user _isn't_ using `mini.icons` but other modules in the entire `mini.nvim` module then this would _force_ the setup 🤔 I suppose we should just check for the global function and hope the user doesn't lazy load it.

echasnovski: > We definitely want to do the `require` at least to protect against the user doing any sort of lazy loading.

> We can leave the `require` and protect users from lazy loading 🎉

I am not really sure what you mean here. My main point is to make sure that the approach here works for users which install whole 'mini.nvim' instead of standalone 'mini.icons' (which is the recommended way) but still want to use 'nvim-web-devicons'.

mehalter: Yeah I totally understand your main point 😄 I just had a few minutes to open up a PR to resolve the issue: https://github.com/stevearc/oil.nvim/pull/441

Thanks for pointing this out!

PR #440, opened 2024-07-06
feat: allow bufnr optional parameter for get_current_dir function

sleeptightAnsiC:
Hi @stevearc, (big fan of your plugin here)

I had a small problem while integrating Oil with a tabline/statusline plugin (lualine.nvim). The problem was that sometimes I had multiple buffers open and Oil buffer wasn't always the currently hovered one. In such cases `get_current_dir` is not working as one would expect as it has a hardcoded reference to buffer `0`. My patch simply adds an optional parameter `bufnr` to said function to make it usable in such cases. This should not break the API in any way and I updated the documentation accordingly. If I screwed something up the PR is open for your edits :)

Cheers!

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -83,10 +83,14 @@ Change how oil determines if the file is hidden
 Toggle hidden files and directories
 
 
-## get_current_dir()
+## get_current_dir(bufnr)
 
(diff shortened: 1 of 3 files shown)


stevearc:
LGTM thanks!

PR #442, opened 2024-07-06
Adding support for moving an existing file to a subdirectory

timtatt:
Previously an oil window with

```
foo.txt
```
changing to: 
```
bar/foo.txt
```
would result in the following error "Filename cannot contain path separator"

With this change, running the above change will create a subdirectory `bar` if it doesn't exist and will move `foo.txt` inside

### DIFF
--- a/lua/oil/mutator/init.lua
+++ b/lua/oil/mutator/init.lua
@@ -93,6 +93,25 @@ M.create_actions_from_diffs = function(all_diffs)
     for _, diff in ipairs(diffs) do
       if diff.type == "new" then
         if diff.id then
+          -- Parse nested files like foo/bar/baz
+          -- Create the parent directories if they don't exist
+          local path_sep = fs.is_windows and "[/\\]" or "/"
+          local pieces = vim.split(diff.name, path_sep)
+          local url = parent_url:gsub("/$", "")
+          for i, v in ipairs(pieces) do
+            if i == #pieces then
+              break
+            end
+
+            url = url .. "/" .. v
+            add_action({
+              type = "create",
+              url = url,
+              entry_type = "directory",
+              link = diff.link,
+            })
+          end
+
           local by_id = diff_by_id[diff.id]
           ---HACK: set the destination on this diff for use later
           ---@diagnostic disable-next-line: inject-field
--- a/lua/oil/mutator/parser.lua
+++ b/lua/oil/mutator/parser.lua
@@ -216,8 +216,6 @@ M.parse = function(bufnr)
         err_message = "No filename found"
       elseif not entry then
         err_message = "Could not find existing entry (was the ID changed?)"
-      elseif parsed_entry.name:match("/") or parsed_entry.name:match(fs.sep) then
-        err_message = "Filename cannot contain path separator"
       end
       if err_message then
         table.insert(errors, {
--- a/tests/mutator_spec.lua
+++ b/tests/mutator_spec.lua
@@ -102,6 +102,32 @@ a.describe("mutator", function()
       }, actions)
     end)
 
+    it("constructs MOVE actions with a new subdirectory", function()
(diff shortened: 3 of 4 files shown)


stevearc:
This currently has the same problems as #400 and #305

```
mkdir foo
touch foo/a.txt
touch a.txt
```

Then nvim . and edit the buffer to look like this:

```
foo/
```

When you save, it moves a.txt to foo/a.txt, obliterating the pre-existing file.

The reason this feature does not exist yet is because of the difficulty in ensuring that these operations are safe and cannot produce unexpected data loss.

timtatt:
> This currently has the same problems as #400 and #305
> 
> ```
> mkdir foo
> touch foo/a.txt
> touch a.txt
> ```
> 
> Then nvim . and edit the buffer to look like this:
> 
> ```
> foo/
> foo/a.txt    // renamed from a.txt
> ```
> 
> When you save, it moves a.txt to foo/a.txt, obliterating the pre-existing file.
> 
> The reason this feature does not exist yet is because of the difficulty in ensuring that these operations are safe and cannot produce unexpected data loss.

Great points, I've pushed a couple ideas to resolve this. Added a `file_exists` method to the adapter which can be used in the parse phase. Currently only implemented for 'files' adapter due to complexity with callback for SSH (open to ideas here). 

Below are the use cases I've identified, let me know if I'm missing any. If you're happy with the approach I'll add some tests to tie it up.

1. When foo/bar.txt exists AND bar.txt is renamed to foo/bar.txt

```
mkdir foo
touch foo/bar.txt
```

```
foo/
foo/bar.txt // renamed from bar.txt
```

Current Behaviour:
image

New Behaviour:
image


2. When foo/bar.txt exists AND bar.txt is renamed to bar/bar.txt AND foo/ is renamed to bar/
 - Expect to throw an error that you cannot create a file in a subdirectory while renaming the same directory

```
mkdir foo
touch foo/bar.txt
touch bar.txt
```

```
bar/ // renamed from foo/
bar/bar.txt // renamed from bar.txt
```




3. When foo/bar.txt exists AND bar/bar.txt is created AND foo/ is renamed to bar/

3a. bar/bar.txt is below bar/
```
mkdir foo
touch foo/bar.txt
```

```
bar/ // renamed from foo/
bar/bar.txt // added
```

Previous Behaviour:
No actions, the bar/bar.txt file is ignored

New Behaviour:
image

3b. bar/bar.txt is above bar/
```
mkdir foo
touch foo/bar.txt
```

```
bar/bar.txt // added
bar/ // renamed from foo/
```

Previous Behaviour:
If bar/bar.txt is before bar/ in the buffer, it will result in:
image

New Behaviour:
image

timtatt:
@stevearc what do you think this solution?

stevearc (changes requested):
Looks like the right general idea of checking for conflicts before performing the actions. Two issues with the current approach:

1. It's not supported in the ssh adapter. It would be ideal if _all_ the adapter special-case logic could be removed, but that may not be possible because the trash adapters may still need to be handled differently. Haven't thought about that too deeply yet.
2. It checks for collisions with the current state of the filesystem, but not the target state. For example: if you put a file into `foo/bar/baz.txt` from multiple locations in the same operation, we won't catch that and they'll clobber each other with only one winning. Likewise, if you put a file into `foo/bar.txt` (with a conflicting existing file) and rename the current file to `foo/baz.txt` in the same operation, it will incorrectly block you.

To be a truly safe operation, each of these nested files will need to be checked to see if there are any conflicts in _any_ open oil buffers, as well as the target directory. Note that the current implementation seems to support `../foo.txt`, which can get complicated. For example, in `/home/user` you can move a file to `../../foo.txt` and `../../../foo.txt`, which _look_ different but both resolve to `/foo.txt`

I suspect that pursuing the above goal will also solve the issue with the unsupported ssh adapter, since the approach won't be using a direct "does this file exist" check.

PR #448, opened 2024-07-17
fix: Handle users and groups with spaces correctly in SSH

annagrram:
This is a fix for a problem I had when using this plugin to SSH our servers at work.
We have group names containing spaces, which makes the line parsing break and raise an exception.

My solution was to work directly with the IDs of the users and groups.

Please let me know if there's a better or more appropriate solution, and I'll be happy to amend this PR.

### DIFF
--- a/lua/oil/adapters/ssh/connection.lua
+++ b/lua/oil/adapters/ssh/connection.lua
@@ -156,7 +156,7 @@ function SSHConnection.new(url)
   else
     self.jid = jid
   end
-  self:run("whoami", function(err, lines)
+  self:run("id -u", function(err, lines)
     if err then
       vim.notify(string.format("Error fetching ssh connection user: %s", err), vim.log.levels.WARN)
     else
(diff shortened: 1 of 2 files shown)


stevearc:
I'm fine with this change. Can you confirm that you've tested:
1. permissions work on allowed directories (i.e. a directory that you have user or group write permissions on is modifiable)
2. permissions work on disallowed directories (a directory you have no write permissions for is not modifiable)

annagrram:
Yeah.
I tested it again today, and it seems to work.

annagrram:
As a side note, I was about to test it on another server but got some exceptions.
The exceptions are related to locale settings that give unexpected results when trying to parse the `ls` line (all of my problems with the SSH adapter seem to originate from this `ls` line parsing).
I have a minor fix for it.
Would you prefer a new PR, or should I add it to this PR?

stevearc:
Great, thanks for checking!

I'd say put that fix in a new PR, just so I can go ahead and merge this one.

PR #451, opened 2024-07-19
feat: allow for toggling of gitignored files

fmgornick:
Hi! i'm a big fan of this plugin and thought this feature would be a nice addition, so i figured it'd be a good opportunity to try and contribute!

I saw there was a recipe provided for hiding gitignored files in an oil buffer and found that it was almost what i wanted.  I personally prefer seeing them by default, but then toggling them off when they get in the way, so i thought it'd be nice if there was a toggle option similar to how `toggle_hidden()` works.

this change adds the ability to toggle on and off gitignored files with the default keymap `gI` (this command does nothing if not in a git repo btw).  I'm not partial to `gI` or anything tho, so if you have a better idea for a default keymap i can always change it!

I also updated the docs accordingly and removed the aforementioned recipe since it's now a config option.  I totally understand if this change seems too invasive, so no hard feelings if you wanna close it out!

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ require("oil").setup({
     ["gs"] = "actions.change_sort",
     ["gx"] = "actions.open_external",
     ["g."] = "actions.toggle_hidden",
+    ["gI"] = "actions.toggle_ignored",
     ["g\\"] = "actions.toggle_trash",
   },
   -- Set to false to disable all of the above keymaps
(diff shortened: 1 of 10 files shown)


stevearc:
I think I'd prefer to keep this as a recipe that lives in the user config. I'd like the built-in logic to be pretty simple so it's easy to reason about adding customizations on top of it

fmgornick:
makes sense!

PR #455, opened 2024-07-22
fix: Force standard C locale when getting `ls` input for parsing in SSH

annagrram:
The locale on the remote machine might result in an unexpected `ls` line, which crashes when parsing.
Therefore, forcing a known locale when executing `ls` for parsing might be better.


stevearc:
LGTM thanks!

PR #456, opened 2024-07-22
fix: add compatibility for Lua 5.1

juru1234:
Remove `goto`'s to be compatible with Neovim built without LuaJIT.

### DIFF
--- a/lua/oil/mutator/parser.lua
+++ b/lua/oil/mutator/parser.lua
@@ -192,115 +192,119 @@ M.parse = function(bufnr)
       seen_names[name] = true
     end
   end
+
   for i, line in ipairs(lines) do
-    if line:match("^/%d+") then
-      -- Parse the line for an existing entry
-      local result, err = M.parse_line(adapter, line, column_defs)
-      if not result or err then
-        table.insert(errors, {
-          message = err,
-          lnum = i - 1,
-          end_lnum = i,
-          col = 0,
-        })
-        goto continue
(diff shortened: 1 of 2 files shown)


lua/oil/mutator/parser.lua
```diff
@@ -203,10 +207,10 @@ M.parse = function(bufnr)
           end_lnum = i,
           col = 0,
         })
-        goto continue
+        return
```
stevearc, line 210: This doesn't seem right. Since the function wraps the for-loop I believe the `return` will be functioning like a `break`, not a `continue`

lua/oil/view.lua
```diff
@@ -630,7 +634,7 @@ local function render_buffer(bufnr, opts)
       jump_idx = #line_table
       M.set_last_cursor(bufname, nil)
     end
-    ::continue::
+   end)()
```
stevearc, line 637: Since this loop doesn't have that many lines, I'd prefer to just nest it under `if M.should_display(entry[FIELD_NAME], bufnr) then` instead of wrapping it in a function

stevearc (changes requested):
Left some feedback

juru1234:
> Left some feedback

Sorry for the wrong code. It passed my small test but wasn't correct. Force pushed another version.

juru1234:
Do I need to push again using `stylua` or will this be applied automatically?

stevearc:
LGTM thanks for the PR!

PR #459, opened 2024-07-27
feat: hidden highlights

### LINKED ISSUE #329: feature request: New highlight group for hidden files and directories

### Describe the feature

I would appreciate it if hidden files and directories had their own highlight group, so they would stand out more.

### Provide background

I got used to seeing the hidden files and directories as "dimmed" in neo-tree, which made it simpler for me to scan through the directory tree.

nice to have

### Additional details

_No response_

zaucy:
simple implementation to add hidden highlights to oil entries

closes #329

## example

image


mehalter:
This would be an awesome addition! I think one improvement I could suggest is also adding a highlight for a default "Hidden" highlight, so like `OilHidden` which links to `Comment` and then each of the other "Hidden highlight groups default link to that. Then it's easier to change the general case as well as have the ability to dive into more specific highlights!

zaucy:
good idea. added!

stevearc:
Thanks for the PR!

Issue #398, opened 2024-06-01
bug: `win_options` gets applied for the first opened buffer/window


NStefan002:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

0.10.0

### Operating system/version

Ubuntu 22.04.4 LTS

### Describe the bug

When starting the new neovim instance with `nvim <dir_name>` and instantly using something like `<C-o>`, `<C-i>` or harpoon jump the first opened buffer has no `signcolumn` nor `number` or `relativenumber` (basically `win_options` get applied to the first buffer/window opened after Oil loads). 

https://github.com/stevearc/oil.nvim/assets/100767853/b825d6ab-cb3f-4a6c-9cfb-11d207a8c193

In the video, you can see that I press `<C-o>` right as the Oil is loading.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. `nvim -u repro.lua dir/`
2. instantly press `<C-o>` (if the jumplist is empty just open up a file and press `G`, that should create one entry in the jumplist and then redo steps 1 and 2)

### Expected Behavior

Buffer/window opened using the mentioned methods should have `statuscolumn`, `number`, and `relativenumber` (if they're set in the neovim config directory, of course).

### Directory structure

dir/1.lua

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
                win_options = {
                    number = false,
                    relativenumber = false,
                    signcolumn = "yes",
                },
            })
        end,
  },
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

[...]

PR #469, opened 2024-08-27
Add check for filetype before setting win_options

csponge:
Fixes an issue where `win_options` were being applied when opening a file in the jumplist at startup.

Ref: https://github.com/stevearc/oil.nvim/issues/398


stevearc:
Thanks for the PR!

PR #470, opened 2024-08-28
doc(recipes): improve git integrated hidden files recipe

mehalter:
I have been using the `is_hidden_file` filter function for adding git integration relatively similar to the recipe that already exists. This makes a couple improvements, first it increases the performance of the `git ls-files` call by removing the need to iterate over a list of files each time, it also adds integration for showing tracked hidden files along with hiding ignored non-hidden files. I've been using this a bit and it has worked great even in huge repos. Let me know what you think!

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -30,47 +30,72 @@ require("oil").setup({
 })
 ```
 
-## Hide gitignored files
+## Hide gitignored files and show git tracked hidden files
 
 ```lua
-local git_ignored = setmetatable({}, {
-  __index = function(self, key)
-    local proc = vim.system(
-      { "git", "ls-files", "--ignored", "--exclude-standard", "--others", "--directory" },
-      {
+-- helper function to parse output
+local function parse_output(proc)
+  local result = proc:wait()
+  local ret = {}
+  if result.code == 0 then
+    for line in vim.gsplit(result.stdout, "\n", { plain = true, trimempty = true }) do
+      -- Remove trailing slash
+      line = line:gsub("/$", "")
+      ret[line] = true
+    end
+  end
+  return ret
+end
+
+-- build git status cache
+local function new_git_status()
+  return setmetatable({}, {
+    __index = function(self, key)
+      local ignore_proc = vim.system(
+        { "git", "ls-files", "--ignored", "--exclude-standard", "--others", "--directory" },
+        {
+          cwd = key,
+          text = true,
+        }
+      )
+      local tracked_proc = vim.system({ "git", "ls-tree", "HEAD", "--name-only" }, {
         cwd = key,
         text = true,
+      })
+      local ret = {
+        ignored = parse_output(ignore_proc),
+        tracked = parse_output(tracked_proc),
       }
-    )
-    local result = proc:wait()
-    local ret = {}
-    if result.code == 0 then
-      for line in vim.gsplit(result.stdout, "\n", { plain = true, trimempty = true }) do
-        -- Remove trailing slash
-        line = line:gsub("/$", "")
-        table.insert(ret, line)
-      end
-    end
 
-    rawset(self, key, ret)
-    return ret
-  end,
-})
+      rawset(self, key, ret)
+      return ret
+    end,
+  })
+end
+local git_status = new_git_status()
+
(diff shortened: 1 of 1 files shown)


stevearc (commented):
What is your frame of reference for "huge repos"? I tried your version in my work repo (root directory, 246 files and directories) and it took 17 seconds to render. Admittedly, the existing version still takes 11 seconds to render, so it's possible that this is just not a feature that can be used by people that work in repos that big 😅

mehalter:
Yeah, I just realized that when I was testing I had not really tested it in a real "large" repo. I get the same performance with this than the meta-table if you just do the hiding of `gitignore` files. When you add in the unhiding of tracked `.` files then it definitely gets slower. Yeah, it's definitely just not very performant. I wonder if there is a good way to build some caching. This type of feature is super handy to be able to really just see git tracked files

stevearc:
What about this?
```lua
local function parse_output(proc)
  local result = proc:wait()
  local ret = {}
  if result.code == 0 then
    for line in vim.gsplit(result.stdout, "\n", { plain = true, trimempty = true }) do
      -- Remove trailing slash
      line = line:gsub("/$", "")
      ret[line] = true
    end
  end
  return ret
end

local git_status = setmetatable({}, {
  __index = function(self, key)
    local ignore_proc = vim.system(
      { "git", "ls-files", "--ignored", "--exclude-standard", "--others", "--directory" },
      {
        cwd = key,
        text = true,
      }
    )
    local tracked_proc = vim.system({ "git", "ls-tree", "HEAD", "--name-only" }, {
      cwd = key,
      text = true,
    })
    local ret = {
      ignored = parse_output(ignore_proc),
      tracked = parse_output(tracked_proc),
    }

    rawset(self, key, ret)
    return ret
  end,
})

require("oil").setup({
  view_options = {
    is_hidden_file = function(name, _)
      local dir = require("oil").get_current_dir()
      -- dotfiles are considered hidden unless tracked
      if vim.startswith(name, ".") then
        return not git_status[dir].tracked[name]
      end
      -- if no local directory (e.g. for ssh connections), always show
      if not dir then
        return false
      end
      -- Check if file is gitignored
      return git_status[dir].ignored[name]
    end,
  },
})
```

mehalter:
Could do something like this:

```lua
local function parse_output(proc)
  local result = proc:wait()
  local ret = {}
  if result.code == 0 then
    for line in vim.gsplit(result.stdout, "\n", { plain = true, trimempty = true }) do
      -- Remove trailing slash
      line = line:gsub("/$", "")
      ret[line] = true
    end
  end
  return ret
end

local function new_git_status()
  return setmetatable({}, {
    __index = function(self, key)
      local ignore_proc = vim.system(
        { "git", "ls-files", "--ignored", "--exclude-standard", "--others", "--directory" },
        {
          cwd = key,
          text = true,
        }
      )
      local tracked_proc = vim.system({ "git", "ls-tree", "HEAD", "--name-only" }, {
        cwd = key,
        text = true,
      })
      local ret = {
        ignored = parse_output(ignore_proc),
        tracked = parse_output(tracked_proc),
      }

      rawset(self, key, ret)
      return ret
    end,
  })
end
local git_status = new_git_status()

local refresh = require("oil.actions").refresh
local orig_refresh = refresh.callback
refresh.callback = function(...)
  git_status = new_git_status()
  orig_refresh(...)
end

require("oil").setup {
  view_options = {
    is_hidden_file = function(name, _)
      local dir = require("oil").get_current_dir()
      -- if no local directory (e.g. for ssh connections), just hide dotfiles
      if not dir then return vim.startswith(name, ".") end
      -- dotfiles are considered hidden unless tracked
      if vim.startswith(name, ".") then
        return not git_status[dir].tracked[name]
      else
        -- Check if file is gitignored
        return git_status[dir].ignored[name]
      end
    end,
  },
}
```

stevearc:
LGTM thanks!

bluebrown:
I tried using this code it didnt do anything. Then I added the blow code
to my `is_always_hidden`, field. It seems to work fine. 

```lua
is_always_hidden = function(name, _)
    vim.fn.system("git check-ignore -q " .. name)
    if vim.v.shell_error == 0 then return true end 
end
```

PR #471, opened 2024-08-28
perf(view): avoid running `is_hidden_file` when `show_hidden` is set

mehalter:
When `show_hidden` is set, we have no need to run `is_hidden_file`. This changes the order of the boolean logic so that the `show_hidden` value can short circuit the evaluation. This causes a significant performance increase when setting `show_hidden`.


stevearc (approved):
LGTM

Out of curiosity, are you encountering performance issues? Roughly how many files are in your directories, and how long is it taking to render?

Issue #462, opened 2024-08-12
feature request: disable LSP file methods


chnn:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

I would love to turn off the LSP file methods completely.

### Provide background

 I work in a large repository where those operations are too slow to ever complete successfully. Trying to complete them sometimes leaves the file tree in a chaotic messy state. In the worst case it has caused me to lose a file altogether.

### What is the significance of this feature?

nice to have

### Additional details

For now I'm setting `timeout_ms=1`.

csponge:
I've made some modifications locally, and I think I got something working. My modifications just adds an `enable` option to the existing `lsp_file_methods`. When disabled, it won't perform the lsp actions.

Before I would do a pull request though, I had a question. Would it be possible to just disable these capabilities for the project you are working on?

- `workspace.fileOperations.willCreate`
- `workspace.fileOperations.willRename`
- `workspace.fileOperations.willDelete`

I think it would be easier to just have a flag in the oil config, but I thought I would pose the question first.

Any thoughts?

chnn:
Yeah, I think either disabling the capabilities or adding the oil config flag would work. The oil config flag is definitely easier!

csponge:
Hi @chnn! I've pushed the changes to my branch. Would you mind pulling my version and giving it a test?

https://github.com/csponge/oil.nvim/tree/feat/disable-lsp-file-methods

Let me know if that fixes things for you.

chnn:
@csponge Gave it a whirl, seems to be working great! I'll continue testing it in day-to-day work

bls-ibf:
since #477 was merged, can we close this issue?

bew:
since #477 was merged, can we close this issue?

PR #477, opened 2024-09-12
Feat/disable lsp file methods

csponge:
This PR adds the option to enable or disable LSP file methods.

ref: https://github.com/stevearc/oil.nvim/issues/462


stevearc:
Thanks for the PR! Did a tiny refactor because I prefer `enabled` over `enable`

Issue #382, opened 2024-05-19
show relative path in window title rather than absolute path


santhosh-tekuri:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

show relative path in window title rather than absolute path

I am using floating window. and absolute path for my project is very long and gets truncated in window title
it would be better if there is an option to show relative path from current working directory in window title

### Provide background

trying to show relative path in window title rather than absolute path when shown in floating mode

### What is the significance of this feature?

nice to have

HarimbolaSantatra:
Absolute path is also very long on status line

rtalexk:
With the merged PR (#482) now you can add the following to your config to achieve that:

```lua
{
  'stevearc/oil.nvim',
  opts = {
    float = {
      get_win_title = function(winid)
        local cwd = vim.fn.getcwd()
        local buf = vim.api.nvim_win_get_buf(winid)
        local file_path = vim.api.nvim_buf_get_name(buf)
        local relative_path = file_path:sub(#cwd - 1)

        -- leading and trailing spaces just as a personal preference
        return ' ' .. relative_path .. ' '
      end,
    },
  },
}
```

santhosh-tekuri:
the above code does not work as expected. since `file_path` is in the format `oil:///.....`

PR #482, opened 2024-09-15
Adding Customisability for floating window title.

PhilippOesch:
Add support to display the window title relative to the current working directory. See #382

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -127,7 +127,7 @@ local default_config = {
     win_options = {
       winblend = 0,
     },
-    -- optionally override the oil buffers window title with custom function: fun(path: string): string
+    -- optionally override the oil buffers window title with custom function: fun(winid: integer): string
     get_win_title = nil,
     -- preview_split: Split direction: "auto", "left", "right", "above", "below".
     preview_split = "auto",
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -129,6 +129,8 @@ local default_config = {
     },
     -- preview_split: Split direction: "auto", "left", "right", "above", "below".
     preview_split = "auto",
+    -- display window title path relative to the current working directory.
+    relative_win_title = false,
     -- This is the config that will be passed to nvim_open_win.
     -- Change values here to customize the layout
     override = function(conf)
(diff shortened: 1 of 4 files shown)


lua/oil/util.lua
```diff
@@ -429,7 +439,7 @@ M.add_title_to_win = function(winid, opts)
         if vim.api.nvim_win_get_buf(winid) ~= winbuf then
           return
         end
-        local new_title = get_title()
+        local new_title = M.get_title()
```
stevearc, line 442: Should pass the winid here

lua/oil/util.lua
```diff
@@ -360,28 +360,38 @@ M.is_floating_win = function(winid)
...
+  local title = vim.api.nvim_buf_get_name(src_buf)
+  local scheme, path = M.parse_url(title)
+
+  if config.adapters[scheme] == "files" then
+    assert(path)
+    local fs = require("oil.fs")
+
+    if config.float.get_win_title ~= nil then
+      return config.float.get_win_title(fs.posix_to_os_path(path))
+    end
```
stevearc, line 377: I think that to make this more general, the config `get_win_title` function should take a winid and return a string. Then we could move this check-return to the top of the function.

lua/oil/config.lua
```diff
@@ -332,6 +334,7 @@ local M = {}
 ---@field max_height integer
 ---@field border string|string[]
 ---@field win_options table<string, any>
+---@field get_win_title fun(path: string): string
```
stevearc, line 337: You'll also need to add an entry on line 347 below. This is for the internal typed representation of the config, below is for the type that is passed in to `oil.setup()`, so is the same except everything is optionally nil.

stevearc (changes requested):
This is a reasonable extension to the configurability. Left a couple notes for small changes.

lua/oil/config.lua
PhilippOesch: Thanks, adjusted.

lua/oil/util.lua
PhilippOesch: Thank you for the response. Implemented the feedback.

lua/oil/util.lua
PhilippOesch: Adjusted.

stevearc:
Looks good, thanks for the PR!

PR #484, opened 2024-09-19
fix: only map ~ for normal mode

staticssleever668:
Allows to switch character case with ~ (tilde) in visual mode while preserving existing ~ :tcd functionality.
Related to [1].

[1]: https://github.com/stevearc/oil.nvim/issues/397 "bug: ~ not respected"


stevearc:
Hmmm...it really was an oversight to use `~` as a default binding at all. Probably should have chosen something else, or left it out of the defaults. Thanks for the PR, I think this improves it!

PR #485, opened 2024-09-19
fix: stick buffer to floating window

### LINKED ISSUE #332: bug: oil.nvim buffer not fix to floating window

- [X] I have read the docs

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2597+gd326e0486

### Operating system/version

linux 6.1

### Describe the bug

When opened as floating window you can hit by accident `ctrl-o` or `ctrl-i` and the floating window will change to the previous/next buffer in the jumplist, which is somewhat annoying.

Consider using `:h winfixbuf` option

minor (annoyance)

### Steps To Reproduce

1. open oil.nvim as floating window
2. hit `ctrl-o` or `ctrl-i`

### Expected Behavior

I would expect the floating window not to change

### Directory structure

_No response_

### Repro

```Lua
not needed
```

Vinzent03:
Set the [winfixbuf](https://neovim.io/doc/user/options.html#'winfixbuf') option for the float window to prevent accidentally switching the buffer. The option has to be partially turned off though to allow switching the buffer. Using the `edit` cmd with `!` would also ignore the `winfixbuf`, but I didn't want to risk any side effects of the bang.

close #332

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -332,6 +332,9 @@ M.open_float = function(dir)
   )
 
   vim.cmd.edit({ args = { util.escape_filename(parent_url) }, mods = { keepalt = true } })
+
+  M._fix_window(true, winid)
+
   -- :edit will set buflisted = true, but we may not want that
   if config.buf_options.buflisted ~= nil then
     vim.api.nvim_set_option_value("buflisted", config.buf_options.buflisted, { buf = 0 })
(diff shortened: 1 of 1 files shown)


stevearc:
I don't love this as a solution. The `winfixbuf` option is basically only suited to a single situation: you have opened a special-purpose window and you only ever want one buffer to be displayed within it. What we would want here is more like "open this window and only allow a certain approved subset of buffers to be displayed". Trying to make `winfixbuf` do this is clunky and will lead to issues. For example, if someone opens a floating oil window and then does `:edit .`, they'll get an error message. They can work around it by passing `!`, but that doesn't change the fact that anyone would reasonably expect `:edit .` to work.

If `winfixbuf` is extended to allow some sort of filter function that makes it function as this intends, I would consider adding it. Until then, you could try the more hacky approach in [stickybuf.nvim](https://github.com/stevearc/stickybuf.nvim/).

PR #487, opened 2024-09-25
Feature request: Show parent directory in a split

DriesOlbrechts:
This is a proof of concept to show the parent directory in a split next to the oil buffer.

I was inspired by [triptych](https://github.com/simonmclean/triptych.nvim).
Triptych does not have buffer file editing which is why I'm creating this PR.

I don't have the knowledge/time to develop this properly so I wanted to make a POC to at least show what I would like to see added.

image

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -129,6 +129,7 @@ local default_config = {
     },
     -- preview_split: Split direction: "auto", "left", "right", "above", "below".
     preview_split = "auto",
+	parent_dir_split = "left",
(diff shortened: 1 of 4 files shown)


DriesOlbrechts:
Triptych has no intention of adding oil-like editing any time soon, so adding parent directory split to oil would be a real positive, to at least my workflow.

[https://github.com/simonmclean/triptych.nvim/issues/9#issuecomment-2282826908](Oil/mini.file like file rename/mv)

Issue #349, opened 2024-04-13
feature request: parent directory as a column/vsplit


ariel-frischer:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

Is it possible to have a ranger-like layout? Its nice to see parent directories and child files/dirs in a 3 column layout like ranger. Has anyone implemented something similar?

![ranger-preview](https://github.com/stevearc/oil.nvim/assets/19893395/ee70f159-9f02-4a83-88e0-f270db27a99f)

### What is the significance of this feature?

strongly desired

stevearc:
Technically possible, would require some work. Not a priority for me, but I would be willing to review a PR that does this.

You could also check out [mini.files](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-files.md), which exclusively uses this kind of view.

DriesOlbrechts:
I have made a POC of this functionality in #487

PR #488, opened 2024-09-26
feat: Show parent directory in a split next to oil buffer

DriesOlbrechts:
This PR adds the feature requested in #349.
 
 A split is opened showing the parent directory for the current oil directory.
 This PR is an improvement on #487.
 After looking at the code again I realised I was completely overcomplicating my solution and rewrote it.
 
 I love what you're doing with Oil.nvim and hope my contribution can add some value!


DriesOlbrechts:
@stevearc any reason this isn't being looked at?
I'm happy to receive any feedback!

stevearc (changes requested):
Apologies for the long wait; I haven't written a full accounting of this anywhere yet. The way I like to manage my projects is by maintaining "inbox zero," so zero open _unanswered_ issues/PRs. Some time last year my backlog got so long that it was causing me a lot of stress _and_ killing my motivation to do any work at all, making the problem worse. I decided to declare "bug bankruptcy" and manually zero out my inbox, only responding to new issues & PRs after that cutoff. The good news is that it worked and I've been able to continue maintaining my projects, but the bad news is that it obviously left a lot of threads dangling, which is unfair to people like you that put in a lot of work crafting a PR or writing up a good bug report. So for you and for everyone else affected, I'm sorry, it's not a directed choice to ignore you, it's just load shedding to maintain my sanity. Due to some recent travel, sickness, more sickness, and more travel, my backlog has been creeping back up, but I'm optimistic that I'll be able to zero it out again over the next month or so. And once I've started engaging with a PR I try to give it higher priority, so hopefully this one won't get dropped again.

Now, on to the review:

I've only done a cursory review of the code, but it appears fine. The main area I'd like to focus on is the API & interface. The first incarnation of oil just had a simple directory buffer, which is easy to reason about. You go to whatever window you want, open oil, mess around, then when you open a file it opens in that same window. Then, preview windows were added as an extra option. Now you can open oil in the current window, but also make a split to show the preview window. It's a little harder to reason about / mucks up the UI a bit more, but in a limited capacity and it's aided a _little_ bit by leaning on vim's concept of a preview window. This PR introduces the possibility for _two_ window splits off of the original, which can get messy and hard to reason about. Three windows is also enough to basically take over all visual space on a given tab.

What do you think about instead constraining this functionality specifically to a floating window view? Right now we have `:Oil --float`, what if we made another specific view like `:Oil --miller` that opens a miller column view in floating windows, like what `mini.files` does? The existing floating window view could be reworked to be a subset of this functionality. Would that serve for your use case?

DriesOlbrechts:
@stevearc 
Since it's been 2 weeks and I don't seem to be able to find time to work on this, I suggest maybe tagging this as 'help wanted' or something similar!
If it's still open when I do find time in the future I will pick it up, but I don't want to block anyone else from trying.

I also accidentally closed the PR when reverting the changes, is it possible to reopen this?
Or maybe move it to a new issue that is more fit for the discussed changes.

Alan-John-Byrne:
Hey there guys, is this planning to be looked at again? Would be really cool to have this.

PR #491, opened 2024-10-08
feat: auto-quit vim if oil is closed as last buffer

cdmill:
Added a simple feature that will auto-quit vim if oil is closed and there are no other buffers open. I added an option to `config.lua` to opt-in to this behavior.

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -66,7 +66,7 @@ local default_config = {
     ["<C-h>"] = { "actions.select", opts = { horizontal = true }, desc = "Open the entry in a horizontal split" },
     ["<C-t>"] = { "actions.select", opts = { tab = true }, desc = "Open the entry in new tab" },
     ["<C-p>"] = "actions.preview",
-    ["<C-c>"] = { "actions.close", opts = { exit_if_last_buf = false }, desc = "Close oil, optionally exit vim if oil is closed as last buffer" },
+    ["<C-c>"] = "actions.close",
     ["<C-l>"] = "actions.refresh",
     ["-"] = "actions.parent",
     ["_"] = "actions.open_cwd",
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -53,8 +53,6 @@ local default_config = {
   constrain_cursor = "editable",
   -- Set to true to watch the filesystem for changes and reload oil
   watch_for_changes = false,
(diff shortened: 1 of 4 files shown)


stevearc (changes requested):
I think this is a fine customization, but it should be done by passing in options to the function rather than a config option. The `oil.close` function should take an option table with something like `close_if_last_buf` as the parameter. Then the [close action](https://github.com/stevearc/oil.nvim/blob/1360be5fda9c67338331abfcd80de2afbb395bcd/lua/oil/actions.lua#L130-L133) should be reworked to accept the same options, similar to how [actions.cd](https://github.com/stevearc/oil.nvim/blob/1360be5fda9c67338331abfcd80de2afbb395bcd/lua/oil/actions.lua#L149-L171) accepts options.

cdmill (commented):
I chose `exit_if_last_buf` to make it more clear that it is separate from `oil.close` but if you prefer 'quit' or 'close' than I can resubmit!

lua/oil/init.lua
```diff
@@ -393,7 +393,7 @@ M.open = function(dir)
 end
 
 ---Restore the buffer that was present when oil was opened
-M.close = function()
+M.close = function(exit_if_last_buf)
```
stevearc, line 396: Let's use an options table for this; it's more future-proof. And add a type annotation so that the auto-generated docs will show the parameter

lua/oil/config.lua
```diff
@@ -66,7 +66,7 @@ local default_config = {
     ["<C-h>"] = { "actions.select", opts = { horizontal = true }, desc = "Open the entry in a horizontal split" },
     ["<C-t>"] = { "actions.select", opts = { tab = true }, desc = "Open the entry in new tab" },
     ["<C-p>"] = "actions.preview",
-    ["<C-c>"] = "actions.close",
+    ["<C-c>"] = { "actions.close", opts = { exit_if_last_buf = false }, desc = "Close oil, optionally exit vim if oil is closed as last buffer" },
```
stevearc, line 69: Let's not change the default close action. The option will be automatically documented in `:help actions.close`. If we tried to put every parameter in the README just for documentation purposes it would be kind of a mess.

stevearc (changes requested):
Just a couple nits left

cdmill:
Any update on this?

stevearc:
LGTM!

PR #495, opened 2024-10-18
feat: `actions.close` to clear pending motions before closing

chrisbellboy:
Added extra logic to ensure that `actions.close` doesn't close Oil if we're in `operator pending` or `visual` modes, instead it exits those modes with an `<Esc>` call and returns early.

This is to improve the UX as I can't be the only one who constantly accidentally closes Oil 😅 When all I wanted to do was clear an unfinished motion or get out of visual mode, but stay in Oil.

P.s. Long term Oil user, and loving life with it!

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -381,6 +381,14 @@ end
 
 ---Restore the buffer that was present when oil was opened
 M.close = function()
+  local mode = vim.api.nvim_get_mode().mode
+  -- If we're in operator pending or visual modes, we should cancel that operation and return
+  if mode == "no" or mode == "v" or mode == "V" then
+    -- Trigger <ESC> to cancel the operation
+    vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<ESC>", true, true, true), "n", false)
+    return
+  end
+
   -- If we're in a floating oil window, close it and try to restore focus to the original window
   if vim.w.is_oil_win then
     local original_winid = vim.w.oil_original_win


stevearc:
Hmmm...I would expect that this could be handled by setting the `mode` of the keymap. What is your current keymap config? What combination of keys are you pressing that winds up with oil closing unexpectedly?

chrisbellboy:
For Oil I've only got this:

`			keymaps = {
				["<Esc>"] = { callback = "actions.close", mode = "n" },
			},
`

But even though that is the case, if you do this:

1. Open Oil
2. Press Ctrl + V to go into Visual Mode. Or press a single `y` to go into Operator Pending mode.
3. Press Escape

Oil will remain open, but the mode gets changed back to Normal mode. You have to hit Escape again to close Oil. Which feels correct to me.

If you do that again but press Ctrl + C instead of Escape in Step 3, then Oil is instantly closed although the mode was also changed back to Normal mode again.

I use Ctrl + C a lot instead of Esc, so it may be a me problem. But in the context of an Oil buffer, it feels like Ctrl + C should behave like Escape. 🤔

stevearc:
@chrisbellboy I recently made some slight modifications to the default keymaps to put more of them as normal-mode-only. Can you check to see if you have any issues on the latest master?

chrisbellboy:
@stevearc Nice one, thanks! I just tested it, and it's really close but it's a bit buggy at the moment: 

The TL;DR is that Ctrl + C now only closes Oil if in normal mode (great). But it can require multiple `Ctrl + C` presses to clear modes to go back to normal mode when in Visual/Operator Pending modes inside Oil. Which Escape doesn't need.

With Oil open:

* If you `Ctrl + V` on a single line and then `Ctrl + C`, Oil now stays open (great), but the Visual mode doesn't get fully cleared. It stays in Visual mode and the operator pending command changes to `1`.
* If you `Ctrl + V` multiples lines, then the operator pending command automatically changes to `3` (or whatever number) and in this situation now hitting `Ctrl + C` does clear it all whilst keeping Oil open.

Similar for operator commands, with Oil open:

* If you hit `y` and then `Ctrl + C`, Oil stays open, but the `y` operator pending remains active. If you hit `Ctrl + C` a second time, now the operator gets cleared.
* If you hit `y3` and then `Ctrl + C`, Oil stays open and the operator pending mode gets cleared correctly.

Thanks again for helping with this!

stevearc:
That's odd, I'm not seeing that behavior. For example, if I do `y` -> `<C-c>` -> `<C-c>` oil closes as expected. Can you provide a minimal config (such as the template in the bug report form) that reproduces this?

PR #497, opened 2024-10-24
fix: pass opts to actions.preview

vex9z7:
It looks like an unintened mistake to not pass the opts to the callback, because without it, the preview window cannot be a horizontal split.

### DIFF
--- a/doc/oil.txt
+++ b/doc/oil.txt
@@ -527,6 +527,12 @@ preview                                                          *actions.previe
     Open the entry under the cursor in a preview window, or close the preview
     window if already open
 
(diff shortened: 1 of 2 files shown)


stevearc (changes requested):
You will also want to add a `parameters` table to the preview action, like the other actions that accept options. Example:
https://github.com/stevearc/oil.nvim/blob/1360be5fda9c67338331abfcd80de2afbb395bcd/lua/oil/actions.lua#L22-L43

stevearc:
Thanks for the PR!

PR #502, opened 2024-10-28
feat: add highlight for orphan links

### LINKED ISSUE #501: feature request: Add a highlight for orphan link

### Describe the feature

It would be helpful if we could set a highlight for orphan links, like LS_COLORS' `or` option.

### Provide background

_No response_

nice to have

### Additional details

_No response_

Foo-x:
Added a highlight for orphan links.

Closes #501

### DIFF
--- a/doc/oil.txt
+++ b/doc/oil.txt
@@ -599,10 +599,13 @@ OilLink                                                               *hl-OilLin
     Soft links in an oil buffer
 
 OilOrphanLink                                                   *hl-OilOrphanLink*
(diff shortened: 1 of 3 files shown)


stevearc:
There have been a few requests for more highlight customization. I want to rethink the highlighting system holistically instead of adding patches for specific cases, so I will be leaving this pending until I can find the time to think about this some more.

stevearc:
Thanks for the PR! I also added a Hl group for the orphaned link's target and highlighted it as an error by default.

PR #504, opened 2024-10-28
feat: add trailing slash on yank_entry

### LINKED ISSUE #503: feature request: Add a trailing slash if the entry type is directory on yank_entry

### Describe the feature

Currently `yank_entry` action does not add a trailing slash even if the entry type is directory.
It would be better to make it consistent with oil buffers.

### Provide background

_No response_

nice to have

### Additional details

_No response_

Foo-x:
Changed to add a trailing slash if the entry type is directory on yank_entry.

Closes #503

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -367,8 +367,8 @@ M.yank_entry = {
       return
     end
     local name = entry.name
-    if entry.type == 'directory' then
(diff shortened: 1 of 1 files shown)


Foo-x:
Fixed it according to the error message.

stevearc:
Thanks for the PR!

PR #508, opened 2024-10-30
feat: highlight config

### LINKED ISSUE #402: feature request: custom highlights

### Describe the feature

In cases like these, it would be nice to be able to define some custom highlight groups, perhaps something like this:

```lua
function has_ts_source(bufnr, name)
  --impl left to the imagination
end

require'oil'.setup {
  -- ...
  view_options = {
    get_custom_highlight = function(name, bufnr)
      if name:gmatch'(.js|.map|.d.ts)$' and has_ts_source(name, bufnr) then
        return { 'OilArtifact' }
      end
    end,
  },
}

vim.cmd.highlight('link OilArtifact Comment')
```

### Provide background

When working in a TypeScript project, it is common to have some build artifacts generated for each source file:

```
pf-icon.e2e.d.ts
pf-icon.e2e.js
pf-icon.e2e.js.map
pf-icon.e2e.ts
pf-icon.spec.d.ts
pf-icon.spec.js
pf-icon.spec.js.map
pf-icon.spec.ts
```

In this case, the user wants the `.d.ts`, `.js` and `.map` artifacts to appear in the buffer, but to receive subdued highlighting.

This is not the same as hiding gitignored files, because oftentimes the user still wants to be able to inspect build artifacts (for example, to verify that they were compiled as expected, or to insert debugging code in the runtime files).

nice to have

### Additional details

Similar to #375 
Similar to https://github.com/nvim-neo-tree/neo-tree.nvim/issues/118

Foo-x:
Added a custom highlight feature.

Closes #402

w/o highlight config
image
image

w/ highlight config
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,8 +232,8 @@ require("oil").setup({
       { "type", "asc" },
       { "name", "asc" },
     },
-    -- Return a highlight group name for each entry
-    highlight = function(entry, is_link_target)
+    -- Customize the highlight group for the file name
+    highlight_filename = function(entry, is_hidden, is_link_target, is_link_orphan)
       return nil
     end,
   },
--- a/README.md
+++ b/README.md
@@ -232,6 +232,10 @@ require("oil").setup({
       { "type", "asc" },
       { "name", "asc" },
     },
+    -- Return a highlight group name for each entry
+    highlight = function(entry, is_link_target)
+      return nil
+    end,
   },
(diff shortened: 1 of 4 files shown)


stevearc:
Similar to my comment on #502, this is going to take some more time to think through. The big reason I haven't done something like this yet is the performance impact.

stevearc:
Thanks for the PR!

PR #514, opened 2024-11-11
feat: add `win_options` to `preview_win`

### LINKED ISSUE #509: feature request: Extend preview.win_options to override win_options

### Describe the feature

I now see how `win_options` extends to the preview window. This is good since I like to have things like `number`, `signcolumn`, and `relativenumber` enabled for various reasons. However, I don't need these things in the preview window (all I want to do is preview the content in the most lightweight manner possible). If `preview.win_options` could be refactored to override the higher level `win_options`, one could disable all the bloat when it comes to previewing.

### Provide background

_No response_

nice to have

### Additional details

_No response_

mehalter:
This adds the ability to customize the window options for the preview window for files/directories. This takes a very basic approach, outside of just adding the `win_options` field to the lua type annotations it makes sure to set the `win_options` when creating the `previewwindow` windows. The other wrinkle it accounts for is to make sure to apply the `preview_win.win_options` after applying oil `win_options` if it's within a `previewwindow`. This makes sure it gets the necessary settings such as `conceal` and `concealcursor` while also applying the settings that the user wants for preview windows.

Let me know what you think of this approach!

Resolves #509

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -143,6 +143,8 @@ local default_config = {
     update_on_cursor_moved = true,
     -- Maximum file size in megabytes to preview
     max_file_size_mb = 100,
+    -- Window-local options to use for preview window buffers
+    win_options = {},
   },
   -- Configuration for the floating action confirmation window
   confirmation = {
@@ -326,12 +328,14 @@ local M = {}
 ---@class (exact) oil.PreviewWindowConfig
 ---@field update_on_cursor_moved boolean
 ---@field max_file_size_mb number
+---@field win_options table<string, any>
 
 ---@class (exact) oil.ConfirmationWindowConfig : oil.WindowConfig
 
 ---@class (exact) oil.SetupPreviewWindowConfig
 ---@field update_on_cursor_moved? boolean Whether the preview window is automatically updated when the cursor is moved
 ---@field max_file_size_mb? number Maximum file size in megabytes to preview
+---@field win_options? table<string, any> Window-local options to use for preview window buffers
 
 ---@class (exact) oil.SetupConfirmationWindowConfig : oil.SetupWindowConfig
 
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -551,6 +551,9 @@ M.open_preview = function(opts, callback)
     end
 
     vim.api.nvim_set_option_value("previewwindow", true, { scope = "local", win = 0 })
+    for k, v in pairs(config.preview_win.win_options) do
+      vim.api.nvim_set_option_value(k, v, { scope = "local", win = preview_win })
+    end
     vim.w.oil_entry_id = entry.id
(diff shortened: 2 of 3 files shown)


mehalter:
Hm, not sure what happened with the "failed" test. I checked the logs and it seems to be passing and locally  it seems to pass

lua/oil/config.lua
```diff
@@ -323,6 +323,7 @@ local M = {}
 
 ---@class (exact) oil.PreviewWindowConfig
 ---@field update_on_cursor_moved boolean
+---@field win_options? table<string, any>
```
stevearc, line 326: You'll want this one to omit the `?` and you'll need to also add this to `oil.SetupPreviewWindowConfig`. This is the internal reference for type safety, the other one is for typing the options passed to `setup()`. It's annoying to keep them both in sync, but it's the only way I could think of to get good type safety (and LuaLS completions) in both locations.

stevearc (changes requested):
Hmmm...not sure what's going on with the tests. ~~Does it pass for you locally~~? I've sometimes seen some finicky behavior like this if there are timeouts in an async test.

edit: ah okay yes it does pass locally. Not sure...try rebasing (which you'll have to do for merge conflicts anyway)

lua/oil/config.lua
mehalter: Ah that makes sense, thanks for the guidance here. I updated the commit appropriately

mehalter:
It looks like something is causing it to hang after the tests all pass. I'm not sure what it would be doing after all the tests are done. I'm taking a deeper look now at how the tests are set up. I don't think anything is happening in this PR to make things get funky 🤔

mehalter:
Ah, it looks like something is causing the `win_options_spec.lua` test file to just completely hang

mehalter:
Ah, I found the problem @stevearc it was causing problems because I was a dingus and forgot to update the default `preview_win.win_options` so it was failing to get setup correctly. Should be good now! Sorry for all the noise!

stevearc:
Thanks for the PR!

jamilraichouni:
Hi!

One question to this.
I set

```lua
require("oil").setup({
    preview_win = {
        win_options = {
            foldenable = false,
        },
    }
})
and it works as long as the preview buffer has no focus and I stay in the oil buffer with the list of dir contents.
But when I make the preview buffer the current buffer my previewed code is immediately folded and the win option is overwritten.
Any idea how to avoid that?
I often use the preview to yank stuff I want to put into some other files. In that scenario it is not helpful, that a disabled folding gets enabled when I want to yank/ put code.

Thanks,
Jamil
```

mehalter:
I'm not sure, just want to remove whatever is in your configuration that is setting foldenabled when entering a buffer

kah-ell:
Maybe `noautocmd` and `eventignore` can get you towards your goal.

Issue #285, opened 2024-01-21
bug: alternate not set with `open_float()`


iovis:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

0.9.5

### Operating system/version

macOS 14.2.1

### Describe the bug

Looks to be exactly what was happening for https://github.com/stevearc/oil.nvim/issues/60, but when using the floating window. I confirmed the behavior of what was fixed in that PR still worked, but when using `require("oil").open_focus()` it doesn't seem to set the current buffer as the alternate.

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. nvim -u repro.lua
2. :e somefile1
3. :lua require('oil').open_float() -- somefile1 is set as alternate
4. navigate to another file somefile2, and edit it -- somefile1 is set as alternate
5. :lua require('oil').open_float() -- somefile1 continues to be the alternate, but should be somefile2

### Expected Behavior

1. nvim -u repro.lua
2. :e somefile1
3. :lua require('oil').open() -- somefile1 is set as alternate
4. use oil to navigate to another file somefile2, and edit it -- somefile1 is set as alternate
5. :lua require('oil').open() -- somefile2 is the alternate
6. use oil to navigate to somefile2 again, and edit it -- somefile1 is again alternate

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

[...]

PR #526, opened 2024-12-02
fix: set alternate when using floating windows (#285)

iovis:
This ensures the alternate is set even when using `open_float()`, I copied the code from `restore_alt_buf`, so I'm sure there could be refactor opportunities. Let me know your thoughts!

Fixes: #285


stevearc:
Everything related to the alternate buffer is very tricky. Please add at least one test to cover this case in https://github.com/stevearc/oil.nvim/blob/master/tests/altbuf_spec.lua. Make sure that it fails without your change applied and passes with it.

iovis:
I completely missed the tests, my bad. I'll give it a go tonight!

iovis:
Done, sorry for missing the tests! I confirmed the new test fails in master but passes on my branch.

stevearc (changes requested):
Thanks for adding the test! I fiddled with it a bit on my end, and I think that a simpler fix is to remove the `keepalt = true` on this line.

https://github.com/stevearc/oil.nvim/blob/7a55ede5e745e31ea8e4cb5483221524922294bf/lua/oil/init.lua#L721

That handles the float win case for me, and doesn't seem to impact any of the other tests.

iovis:
> Thanks for adding the test! I fiddled with it a bit on my end, and I think that a simpler fix is to remove the keepalt = true on this line.

You're right, I can confirm it works on my end and it looks like a much more appropriate fix. I just sent the revised code!

stevearc:
Something strange going on with the test failures in 0.8. I'll investigate when I get a chance

iovis:
I think I found the issue in v0.8, when I removed `keepalt` completely this table:

```lua
-- lua/oil/init.lua:717
local mods = {                  
  vertical = opts.vertical,     
  horizontal = opts.horizontal, 
  split = opts.split,           
}                               
```

Could result in an empty table, which apparently neovim v0.8 doesn't like:

```
Error executing vim.schedule lua callback: ~/code/vim/oil.nvim/lua/oil/init.lua:745: 'mods' must be a Dictionary
stack traceback:
        [C]: in function 'cmd'
        ~/code/vim/oil.nvim/lua/oil/init.lua:745: in function 'callback'
        ~/code/vim/oil.nvim/lua/oil/adapters/files.lua:260: in function ''
```

Changing it from removing it completely to `keepalt = false` resolves the issue and the tests still pass (which I cleaned up a bit).

stevearc:
Great, thanks for the fix!

tjex:
Thank you so much for this everyone!!!!

PR #530, opened 2024-12-06
fix: improper file name escaping

### LINKED ISSUE #529: bug: Oil fails with E77 when used in file with weird directory name such as `{a,b}`

- [X] I have read the docs

### Neovim version (nvim -v)

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1713773202 Run "nvim -V1 -v" for more info

### Operating system/version

NixOS 24.11

### Describe the bug

When in a buffer which corresponds to a file in a directory with a name such as `{a,b}` using `:Oil` fails with the following error:

```
Error executing Lua callback: ...ir/pack/myNeovimPackages/start/oil.nvim/lua/oil/init.lua:372: Vim:E77: Too many file names
stack traceback:
	[C]: in function 'edit'
	...ir/pack/myNeovimPackages/start/oil.nvim/lua/oil/init.lua:372: in function <...ir/pack/myNeovimPackages/start/oil.nvim/lua/oil/init.lua:361>
	...ir/pack/myNeovimPackages/start/oil.nvim/lua/oil/init.lua:1137: in function <...ir/pack/myNeovimPackages/start/oil.nvim/lua/oil/init.lua:1093>
```

Note: it seems the directory name must have a `,` since `{a}` works but `{,}` does not.

Seems to be something to do with how `edit` works since `:edit {a,b}/file_name.txt` does not work with the same error E77 but `:exec "edit " .. fnameescape("{a,b}/file_name.txt")` does work.

### Steps To Reproduce

1. Create a directory with a name such as `{a,b}`
2. Create a file in this directory with any name such as `filename.txt`
3. Open the file in Neovim using your preferred method.
4. Type `:Oil` 

### Expected Behavior

Opens the file in the oil directory thingy.

### Directory structure

{a,b}/file_name.txt

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

lucascool12:
Changes the util file escape function to use the builtin `fnameescape` instead which fixes #529.

### DIFF
--- a/lua/oil/util.lua
+++ b/lua/oil/util.lua
@@ -21,7 +21,7 @@ end
 ---@param filename string
 ---@return string
 M.escape_filename = function(filename)
-  local ret = filename:gsub("([%%#$])", "\\%1")
+  local ret = vim.fn.fnameescape(filename)
   return ret
 end
 


stevearc:
Thanks for the fix!

Issue #524, opened 2024-11-29
bug: Oil is hijacking the preview window on open


GustavoKatel:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

v0.10.2

### Operating system/version

MacOS 15.1.1

### Describe the bug

When opening Oil in any window, it hijacks the preview window if it's already open. So for example, if using with dadbod-ui, which uses the preview window to show the db output, oil will take over the preview window.

This is not desirable :( 

I suspect [this function](https://github.com/stevearc/oil.nvim/blob/3c2de37accead0240fbe812f5ccdedfe0b973557/lua/oil/init.lua#L348), is unconditionally checking if the preview window is open.

I think it should only take over the preview window *if* requested by the user, otherwise it will continue to cause unexpected results.

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

1. Open two splits
2. Split 1, run: `:set previewwindow`
3. Split 2, run: `:Oil`

the preview window on split 1 will be taken over by Oil and the contents will be erased

### Expected Behavior

Oil should not open the preview window on open, specially using the `previewwindow`, if not requested

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Did you check the bug with a clean config?

[...]

PR #532, opened 2024-12-08
only use the previewwindow if requested by the user

GustavoKatel:
Tentative fix for https://github.com/stevearc/oil.nvim/issues/524

closes https://github.com/stevearc/oil.nvim/issues/524

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -454,7 +454,7 @@ M.open_preview = function(opts, callback)
     end
   end
 
-  local preview_win = util.get_preview_win()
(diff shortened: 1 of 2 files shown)


stevearc (changes requested):
This is the correct approach to fix the issue, but why did you remove the calls to `update_preview_window`? I'm pretty sure that will lead to out-of-date previews in some cases.

GustavoKatel:
> but why did you remove the calls to `update_preview_window`?

late night coding. my bad :) fixed it

stevearc:
Thanks for the PR!

PR #534, opened 2024-12-10
Fix: filename with newline

jhonnyV-V:
If a file has a newline character in a filename for unknown reason (in my case turns out bun $().text() returns a new line) oil would crash, this should fix the issue, I am not really familiar with the codebase, so I'm not really sure if that is the best way to apply the change, so I'm open to any feedback


stevearc:
Thanks for the report! I've fixed it in a couple different locations

PR #537, opened 2024-12-13
Add config key to set the file creation permissions

fortizc:
This PR defines a new config key called `create_files_mode` to set the default permission used when a file is created.
Currently, Oil uses 644 (420 in decimal) but, at least on Linux most of the tools like (n)vim, touch, or even a simple `> filename` by default uses 664 (436 decimal) so, with this option the user is able to set the default value.

### DIFF
--- a/doc/oil.txt
+++ b/doc/oil.txt
@@ -10,6 +10,7 @@ CONTENTS                                                            *oil-content
   5. Actions                                                     |oil-actions|
   6. Highlights                                               |oil-highlights|
   7. Trash                                                         |oil-trash|
+  8. File Creation                                         |oil-file_creation|
 
 --------------------------------------------------------------------------------
 CONFIG                                                                *oil-config*
@@ -45,6 +46,8 @@ CONFIG                                                                *oil-confi
       },
       -- Send deleted files to the trash instead of permanently deleting them (:help oil-trash)
       delete_to_trash = false,
+      -- Set the default mode to create files (:help oil-file_creation)
+      create_files_mode = 420,
       -- Skip the confirmation popup for simple operations (:help oil.skip_confirm_for_simple_edits)
       skip_confirm_for_simple_edits = false,
       -- Selecting a new/moved/renamed file or directory will prompt you to save changes first
@@ -713,6 +716,36 @@ Mac:
 
 Windows:
     Oil supports the Windows Recycle Bin. All features should work.
+
+--------------------------------------------------------------------------------
+FILE CREATION                                                  *oil-file_creation*
+
+
+Oil defines `create_files_mode` to set the permission for new files. The mode
+follows the Posix numeric standard which use octal notation (base-8), however
+the variable must be set in decimal.
+
+    00700 user (file owner) has read, write, and execute permission
+    00400 user has read permission
+    00200 user has write permission
+    00100 user has execute permission
+    00070 group has read, write, and execute permission
+    00040 group has read permission
+    00020 group has write permission
+    00010 group has execute permission
+    00007 others have read, write, and execute permission
+    00004 others have read permission
+    00002 others have write permission
+    00001 others have execute permission
+    (from the open (2) man pages)
+
+An easy way to provide the right value for `create_files_mode` is using
+`tonumber(e [, base])` function with `base = 8`, for example:
+>lua
+    create_files_mode = tonumber("0644", 8)
+
+The default value is 420 in decimal or 0644 in octal
+
 
 ================================================================================
 vim:tw=80:ts=2:ft=help:norl:syntax=help:
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -28,6 +28,8 @@ local default_config = {
   },
   -- Send deleted files to the trash instead of permanently deleting them (:help oil-trash)
   delete_to_trash = false,
+  -- Set the default mode to create files (:help oil-file_creation)
+  create_files_mode = 420,
   -- Skip the confirmation popup for simple operations (:help oil.skip_confirm_for_simple_edits)
   skip_confirm_for_simple_edits = false,
   -- Selecting a new/moved/renamed file or directory will prompt you to save changes first
@@ -223,6 +225,7 @@ default_config.view_options.highlight_filename = nil
 ---@field buf_options table<string, any>
 ---@field win_options table<string, any>
 ---@field delete_to_trash boolean
+---@field create_files_mode integer
 ---@field skip_confirm_for_simple_edits boolean
 ---@field prompt_save_on_select_new_entry boolean
 ---@field cleanup_delay_ms integer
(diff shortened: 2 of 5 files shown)


lua/oil/config.lua
```diff
@@ -28,6 +28,8 @@ local default_config = {
   },
   -- Send deleted files to the trash instead of permanently deleting them (:help oil-trash)
   delete_to_trash = false,
+  -- Set the default mode to create files (:help oil-file_creation)
+  create_files_mode = 420,
```
stevearc, line 32: We don't expect most users to need to modify this, so let's move it down to below the `view_options`.

lua/oil/config.lua
```diff
@@ -408,6 +412,8 @@ M.setup = function(opts)
   if opts.preview and not opts.confirmation then
     new_conf.confirmation = vim.tbl_deep_extend("keep", opts.preview, default_config.confirmation)
   end
+
+  M.file_mode = opts.create_files_mode
```
stevearc, line 416: Why are we renaming this here?

lua/oil/config.lua
stevearc: If we're adding the ability to adjust the mode of the created files, we should do the same for directories. We could use two options for this `new_file_mode` and `new_dir_mode`.

stevearc: Since most people are used to thinking about file permissions in octal, the config option should be octal and we should automatically convert it.

doc/oil.txt
```diff
@@ -714,5 +717,35 @@ Mac:
...
+    (from the open (2) man pages)
+
+An easy way to provide the right value for `create_files_mode` is using
+`tonumber(e [, base])` function with `base = 8`, for example:
+>lua
+    create_files_mode = tonumber("0644", 8)
+
+The default value is 420 in decimal or 0644 in octal
+
+
```
stevearc, line 749: I actually don't think we need this section. It's explaining:
1. what the octal file permissions mean
1. how to convert octal to decimal
1. the default value

Point 3 is already documented in the options section. Point 2 is obviated if we accept the permission in octal. And for point 1 I would argue that if you don't know what octal file permissions are, you probably shouldn't be changing this config value and I don't think the oil docs are the correct place to be learning about them.

stevearc (changes requested):

doc/oil.txt
fortizc: Ok, then I will remove it

lua/oil/config.lua
fortizc: ok, sounds good

lua/oil/config.lua
fortizc: I will fix it

lua/oil/fs.lua
```diff
@@ -146,7 +147,6 @@ end
 ---@param dir string
 ---@param mode? integer
 M.mkdirp = function(dir, mode)
-  mode = mode or 493
   local mod = ""
```
stevearc, line 150: According to the docs, `uv.fs_mkdir` expects a non-nil integer as the second argument. Did you intend to replace this with the value in the config file?

doc/oil.txt
```diff
@@ -122,6 +122,10 @@ CONFIG                                                                *oil-confi
           return nil
         end,
       },
+      -- Set the default mode to create files
+      new_files_mode = 644,
+      -- Set the default mode to create folders
+      new_folder_mode = 755,
```
stevearc, line 128: I am fully aware of how this sounds, and I'm really sorry to be this pedantic, but "folder" is Windows terminology. Typically in Unix contexts we would refer to these as "directories". Elsewhere in this plugin we always refer to these as directories. For consistency sake, could we rename this to `new_dir_mode`?

stevearc (changes requested):

lua/oil/fs.lua
fortizc: Yes, the idea is to use the config value, and now we have a default one, so it's always available

doc/oil.txt
fortizc: Done

lua/oil/fs.lua
stevearc: Then you'll need to update the function signature to make mode non-optional, and update the call here to pass the mode 
https://github.com/stevearc/oil.nvim/blob/c12fad2d225d8f81fadd48521d253607fe25465c/lua/oil/adapters/ssh.lua#L371

lua/oil/config.lua
```diff
@@ -408,6 +416,9 @@ M.setup = function(opts)
   if opts.preview and not opts.confirmation then
     new_conf.confirmation = vim.tbl_deep_extend("keep", opts.preview, default_config.confirmation)
   end
+
+  new_conf.new_file_mode = tonumber(tostring(new_conf.new_file_mode), 8)
+  new_conf.new_dir_mode = tonumber(tostring(new_conf.new_dir_mode), 8)
```
stevearc, line 421: nit: you don't actually need the `tostring()` here. `tonumber` will convert an integer or a string the same way.

stevearc (changes requested):

lua/oil/config.lua
fortizc: Not sure about it, i try the following code:
```lua
local x = 493
print(tonumber(x, 8))
```
and I get the following error:
```
$ lua test.lua
lua: test.lua:2: bad argument #1 to 'tonumber' (string expected, got number)
stack traceback:
        [C]: in function 'tonumber'
        test.lua:2: in main chunk
        [C]: in ?
```
I'm not a Lua expert, in fact this is my first productive code in Lua, so maybe I'm ignoring something

stevearc: the second argument to [tonumber](https://www.lua.org/manual/5.1/manual.html#pdf-tonumber) is the base to interpret the number as. 493 is not a valid octal number. If you do `tonumber(755, 8)` it prints out `493` as expected

fortizc: Again the same:
```
$ lua -e "print(tonumber(755, 8))"
lua: (command line):1: bad argument #1 to 'tonumber' (string expected, got number)
stack traceback:
        [C]: in function 'tonumber'
        (command line):1: in main chunk
        [C]: in ?
```
Maybe this has something to do with my interpreter? to run this examples I used the standalone interpreter:
```
$ lua
Lua 5.4.6  Copyright (C) 1994-2023 Lua.org, PUC-Rio
```

stevearc: Oh interesting, must be a difference between LuaJIT and Lua then (or 5.1 and 5.4). I retract the nit

fortizc:
Hey, all changes are done, so it's ready if you want to review it again. Thanks!

lua/oil/fs.lua
```diff
@@ -146,9 +147,8 @@ end
 ---@param dir string
 ---@param mode? integer
 M.mkdirp = function(dir, mode)
-  mode = mode or 493
-  local mod = ""
   local path = dir
+  local mod = mode
```
stevearc, line 151: What is going on here? This change looks completely wrong

stevearc (changes requested):
The previous comment about `mkdirp` has not been addressed, and  new bugs were introduced in that function

PR #540, opened 2024-12-20
feat: horizontal/vertical padding

ilyazz-zzayli:
This PR allows the user to specify different padding-values for the left and right sides (horizontal) and for the top and bottom sides (vertical).


stevearc:
Thanks for the PR, but I do not want to merge this. There's a fine balance between adding customization options and making the config options too complex. In this case, you can already achieve this behavior in your own config by using the `float.override` function and changing the row/col/width/height of the window. Because it's already possible (albeit with slightly more effort), I don't think it's worth adding more special cases to the config shape.

PR #550, opened 2025-01-06
test: fix tests shebang for NixOS

abdennourzahaf:
This PR fixes running tests by using a shebang that works for NixOS.


stevearc:
LGTM

PR #551, opened 2025-01-06
Fix custom directory highlights

Landaman:
This PR fixes an issue with custom highlights on directories.

Currently, the custom highlights apply as requested, but remove the trailing '/', causing Oil to consider the directories as deleted and files with the same name as directories to be created.

Simple fix checks if the type is a directory and adds the '/' back


stevearc:
Thanks for the fix!

PR #552, opened 2025-01-06
Add support for bufnr in custom highlights

Landaman:
This PR enables you to determine the full directory path, enabling, for example, HL groups for Git.

Previously this was provided for `is_hidden_file` or `is_always_hidden` but not for `highlight_filename`.

Code changes are simple, since callers of `highlight_filename` always have bufnr anyway.


stevearc:
LGTM thanks!

Issue #554, opened 2025-01-07
bug: modifiable is set off on running AFS filesystem


BozeBro:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

NVIM v0.10.1

### Operating system/version

Ubuntu 22.04.5 LTS

### Describe the bug

When running on my mounted AFS Cell, running nvim on the oil buffer opens it with modifiable off. To work around this, I must call `:set ma` each time to edit the Oil Buffer and write to it. 

I made a PR that fixes the problem by using the uv.fs_access similar to `access(2)` on unix and `GetFileAttributesW()` on Windows that I will link. #555 

The image here shows that the owner of the file is not my own uid, but a placeholder/alternate id of me.
<img width="517" alt="Screenshot 2025-01-07 at 12 47 52 PM" src="https://github.com/user-attachments/assets/63c2b61b-7e3b-4a2b-9038-6d343ce3682a" />

### What is the severity of this bug?

tolerable (can work around it)

### Steps To Reproduce

1. Be in in the AFS Mounted directory
2. Create a test/ directory
3. nvim -u repo.lua
4. :Oil .
5. press o

### Expected Behavior

When opening and writing to the Oil Buffer, modifiable should be on.

### Directory structure

/afs/.../test/example.txt
/afs/.../test/repo.lua

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})
```

[...]

PR #555, opened 2025-01-07
fix: fs_access on dir to check if possible to write to the directory (#554)

BozeBro:


stevearc:
This is great! I didn't know that API existed, but it's definitely what we should be using here. Thanks!

Issue #556, opened 2025-01-11
feature request: copy/paste to system clipboard


stevalkr:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

Hi! This plugin has become an essential tool for me to browse file directories. However, in some cases, using the system’s graphical file explorer (and any other tools) is more convenient. I’d like the ability to copy files from Neovim and paste them into the system clipboard (and vice versa), just like copying and pasting files in a graphical file explorer. 

### Provide background

When I connect a removable device, the path is often difficult to input manually. Similarly, when I want to send files to other applications (like a browser or chat software), I have to manually open the file explorer, copy the file, and then paste it into the application.

### What is the significance of this feature?

strongly desired

### Additional details

I’m not sure if this is feasible on all systems, but on macOS, it can be achieved using [osascript](https://apple.stackexchange.com/questions/15318/how-to-use-terminal-to-copy-a-file-to-the-clipboard).

stevearc:
There are actions to copy the file path, but if you want to copy the file itself that would require something like the osascript you linked. I would review a PR that adds an action with that functionality.

ethanuppal:
This would be pretty cool -- it seems it's a bit [harder on Linux](https://unix.stackexchange.com/questions/53503/copying-files-from-command-line-to-clipboard) but very doable.

ethanuppal:
As a sidenote, the keybinds `gy` and `gp` seem suitable for these actions.

stevalkr:
I’ve created a simple version of copying to the macOS clipboard, and it works. But pasting from clipboard I need more investigation

stevalkr:
`parser.parse` isn’t the best fit for asynchronous file operations. We could add a synchronous `add_file` function to the adapters to support not only files. 

Or, we could implement it during rendering, though that’s not ideal either, but at least it’s asynchronous. If you all have any better ideas, please let me know, and I’d be happy to help.

xzbdmw:
> Similarly, when I want to send files to other applications (like a browser or chat software), I have to manually open the file explorer, copy the file, and then paste it into the application.

I'm using this approach to open the path in finder, after that you can copy it from there (one more step though)
```lua
                ["<c-cr>"] = function()
                    require("oil.actions").yank_entry.callback()
                    local path = vim.fn.getreg('"')
                    -- show in finder
                    vim.system({ "open", "-R", path }):wait()
                end,
                ["<d-cr>"] = function()
[...]

Issue #156, opened 2023-08-19
feature request: pasting path of files into an oil buffer


sassanh:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

It would be nice if oil.nvim could handle absolute file paths pasted into its buffers. In macOS when you copy a file in Finder, you can paste it in terminal and it will input its absolute path, if oil.nvim can copy the files upon writing the buffer, it will be possible to copy from Finder into oil.nvim.

stevearc:
To clarify: if you put an absolute path in an oil buffer and save, you want it to copy the file from the absolute path into the current buffer?

sassanh:
> To clarify: if you put an absolute path in an oil buffer and save, you want it to copy the file from the absolute path into the current buffer?

Yes

sassanh:
For relative paths, I think it may increase the potential for errors. Users may be confused if they are addressing a relative path from current directory of oil.nvim, the cwd of Noevim or the directory of the last opened file, etc.

So we can give up on relative paths altogether or we can consider resolving them to absolute paths upon pasting so that users can see and confirm the resolved path before writing the buffer.

PR #559, opened 2025-01-14
feat: copy/paste to system clipboard

stevalkr:
This PR implements basic features of
- #556 
- #156

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -320,7 +320,7 @@ M.refresh = {
 
 local function open_cmdline_with_path(path)
   local escaped =
-    vim.api.nvim_replace_termcodes(": " .. vim.fn.fnameescape(path) .. "<Home>", true, false, true)
+      vim.api.nvim_replace_termcodes(": " .. vim.fn.fnameescape(path) .. "<Home>", true, false, true)
   vim.api.nvim_feedkeys(escaped, "n", false)
 end
 
@@ -418,6 +418,34 @@ M.copy_entry_filename = {
   end,
 }
 
+M.copy_to_system_clipboard = {
+  desc = "Copy the entry under the cursor to the system clipboard",
+  callback = function()
+    local fs = require("oil.fs")
+    local entry = oil.get_cursor_entry()
+    local dir = oil.get_current_dir()
+    if not entry or not dir then
+      return
+    end
+    local path = dir .. entry.name
+
+    if fs.is_mac then
+      local cmd = "osascript -e 'on run args' -e 'set the clipboard to POSIX file (first item of args)' -e end '%s'"
+      local jid = vim.fn.jobstart(string.format(cmd, path), {
+        stdout_buffered = true,
+        on_exit = function(j, exit_code)
+          if exit_code ~= 0 then
+            vim.notify(string.format("Error copying '%s' to system clipboard", path), vim.log.levels.ERROR)
+          end
+        end,
+      })
+      assert(jid > 0, "Failed to start job")
+    else
+      vim.notify("System clipboard not supported on this platform", vim.log.levels.WARN)
+    end
+  end,
+}
+
 M.open_cmdline_dir = {
   desc = "Open vim cmdline with current directory as an argument",
   deprecated = true,
@@ -455,7 +483,7 @@ M.change_sort = {
           order = order == "ascending" and "asc" or "desc"
           oil.set_sort({
             { "type", "asc" },
-            { col, order },
+            { col,    order },
           })
         end
       )
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -320,7 +320,7 @@ M.refresh = {
 
 local function open_cmdline_with_path(path)
   local escaped =
-      vim.api.nvim_replace_termcodes(": " .. vim.fn.fnameescape(path) .. "<Home>", true, false, true)
+    vim.api.nvim_replace_termcodes(": " .. vim.fn.fnameescape(path) .. "<Home>", true, false, true)
   vim.api.nvim_feedkeys(escaped, "n", false)
 end
 
@@ -432,11 +432,46 @@ M.copy_to_system_clipboard = {
     if fs.is_mac then
       local cmd = "osascript -e 'on run args' -e 'set the clipboard to POSIX file (first item of args)' -e end '%s'"
       local jid = vim.fn.jobstart(string.format(cmd, path), {
-        stdout_buffered = true,
         on_exit = function(j, exit_code)
           if exit_code ~= 0 then
             vim.notify(string.format("Error copying '%s' to system clipboard", path), vim.log.levels.ERROR)
           end
+        end,
+      })
+      assert(jid > 0, "Failed to start job")
+    else
+      vim.notify("System clipboard not supported on this platform", vim.log.levels.WARN)
+    end
+  end,
+}
+
+M.paste_from_system_clipboard = {
+  desc = "Paste the system clipboard into the current oil directory",
+  callback = function()
+    local fs = require("oil.fs")
+    local dir = oil.get_current_dir()
+    if not dir then
+      return
+    end
+
+    if fs.is_mac then
+      local path = nil
+      local cmd = "osascript -e 'on run' -e 'POSIX path of (the clipboard as «class furl»)' -e end"
+      local jid = vim.fn.jobstart(cmd, {
+        stdout_buffered = true,
+        on_stdout = function(j, output)
+          if #output > 1 then
+            path = vim.uv.fs_realpath(output[1])
+          end
+        end,
+        on_exit = function(j, exit_code)
+          if exit_code ~= 0 or path == nil then
+            vim.notify(string.format("Error pasting file path from system clipboard"), vim.log.levels.ERROR)
+            return
+          end
+          local url = "oil://" .. path -- TODO: try to determine the adapter
+          local pos = vim.api.nvim_win_get_cursor(0)
+          vim.api.nvim_buf_set_lines(0, pos[1], pos[1], false, { url })
         end,
       })
       assert(jid > 0, "Failed to start job")
@@ -483,7 +518,7 @@ M.change_sort = {
           order = order == "ascending" and "asc" or "desc"
           oil.set_sort({
             { "type", "asc" },
-            { col,    order },
+            { col, order },
           })
         end
       )
--- a/lua/oil/mutator/parser.lua
+++ b/lua/oil/mutator/parser.lua
@@ -296,12 +296,38 @@ M.parse = function(bufnr)
             name, link = unpack(link_pieces)
           end
           check_dupe(name, i)
-          table.insert(diffs, {
-            type = "new",
-            name = name,
-            entry_type = entry_type,
-            link = link,
-          })
+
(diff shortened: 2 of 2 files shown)


stevalkr:
Hi @stevearc, I’m having a bit of trouble with the `sshFs.load` function. It works great for SSH connections that don’t require passwords, but for those that do, the float window can’t open because `vim.wait` is blocking it. Is there a way to let this function wait for the connection without blocking `vim.api.nvim_open_win`?

lua/oil/adapters/ssh/connection.lua
```diff
@@ -246,8 +264,14 @@ function SSHConnection:_handle_output(start_i)
...
-    elseif last_line:match("Password:%s*$") then
-      self:open_terminal()
+    elseif last_line:lower():match("password:%s*$") then
+      if not self.connected then
+        local pw = vim.fn.inputsecret(string.format("%s password:", url_to_str(self.url)))
+        pcall(vim.api.nvim_chan_send, self.jid, pw .. "\r")
+        pcall(vim.api.nvim_chan_send, self.term_id, "(secret)\r")
+      else
+        self:open_terminal()
+      end
```
stevalkr, line 274: This modification allows inputting passwords during `vim.wait`. However, without implementing additional logic to detect incorrect passwords (Permission denied), the behavior will be weird if the password is incorrect.

stevearc:
This PR looks like it's doing two things: yanking/pasting using osascript, and changing the buffer parsing logic to allow for `oil://` urls. The first is fine, the second, as you have discovered, adds a lot of complexity. Especially this handling of `oil-ssh://` urls; isn't that case impossible to encounter with the copy/paste? The only way it could occur is if people started manually inputting or copying oil urls, which is an entirely new workflow that I'm not convinced we should support.

As an alternative approach to support pasting, what if we found the parent directory of the pasted file, performed an async read of that directory, and then properly inserted the correct line into the current buffer? It's more logic on paste, but it leaves the rest of the internals the same.

stevalkr:
Hi @stevearc! I’ve implemented the method you suggested. I haven’t thoroughly tested all the cases, but it should work fine. However, the code might need some cleaning up :)

stevalkr:
I’ve also added support for Linux (but it needs some testing). But the default file explorer 'nautilus' on Ubuntu doesn’t support the MIME type 'text/uri-list'. I think we could add a config option so the user can specify their own command for reading and writing the clipboard. Also, include an example of supporting 'nautilus’s special protocol in the readme.

copy:
```
echo -en 'copy\\nfile://%s\\0' | xclip -i -selection clipboard -t x-special/gnome-copied-files
``` 

paste:
```
xclip -o -selection clipboard -t x-special/gnome-copied-files | grep --text --color=never file://
```

stevearc (changes requested):
Hmmm...I tested it on my mac but I can't seem to get it working. Copying a file in finder and pasting produces
```
...tevearc/dotfiles/vimplugins/oil.nvim/lua/oil/actions.lua:503: bad argument #2 to 'fs_realpath' (function or callable table expected, got number)
stack traceback:
^I[C]: in function 'fs_realpath'
^I...tevearc/dotfiles/vimplugins/oil.nvim/lua/oil/actions.lua:503: in function <...tevearc/dotfiles/vimplugins/oil.nvim/lua/oil/actions.lua:501>
```
If I copy to the system clipboard, there's no feedback or anything but I can't seem to paste anything into finder. Any ideas?

stevalkr:
> Copying a file in finder and pasting produces
> 
> ```
> ...tevearc/dotfiles/vimplugins/oil.nvim/lua/oil/actions.lua:503: bad argument #2 to 'fs_realpath' (function or callable table expected, got number)
> stack traceback:
> ^I[C]: in function 'fs_realpath'
> ^I...tevearc/dotfiles/vimplugins/oil.nvim/lua/oil/actions.lua:503: in function <...tevearc/dotfiles/vimplugins/oil.nvim/lua/oil/actions.lua:501>
> ```

My bad, should be resolved in the latest commit

> If I copy to the system clipboard, there's no feedback or anything but I can't seem to paste anything into finder. Any ideas?

I added a notify when file is copied to system clipboard, but I can copy and paste as expected. Could you provide me your macOS and shell version? It might be an issue with shell unquoting or a change in osascript

lua/oil/actions.lua
```diff
@@ -418,6 +418,157 @@ M.copy_entry_filename = {
...
+    if not dir or not entry then
+      return
+    end
+    local path = dir .. entry.name
+    local cmd
+    if fs.is_mac then
+      cmd =
+        "osascript -e 'on run args' -e 'set the clipboard to POSIX file (first item of args)' -e 'end run' '%s'"
+    elseif fs.is_linux then
+      cmd = "echo -en '%s\\n' | xclip -i -selection clipboard -t text/uri-list"
```
maxbol, line 436: xclip is only for X11, for wayland you would have to use wl-clipboard, right?

stevalkr: Sure! Could you help me find and test the equivalent command using `wl-clipboard`? I’m not using Linux anymore, so it’s difficult for me to test these things out.

maxbol: I'm on MacOS unfortunately, just spotted this and thought I'd let you know :-)

stevearc:
I did some refactoring and also fixed some problems with copy/paste on KDE. I tested and it still appears to work on Mac. Will test on Gnome when I get a chance. Take a look to make sure the changes look reasonable, and test to make sure it still works on your machine(s) as well.

lua/oil/actions.lua
```diff
@@ -476,14 +495,32 @@ M.paste_from_system_clipboard = {
...
     elseif fs.is_linux then
-      cmd = "xclip -o -selection clipboard -t text/uri-list"
+      local xdg_session_type = vim.env.XDG_SESSION_TYPE:lower()
+      if xdg_session_type:find("x11") then
+        if
+          vim.env.XDG_SESSION_DESKTOP:lower():find("gnome")
+          or vim.env.XDG_CURRENT_DESKTOP:lower():find("gnome")
+        then
+          cmd =
+            "xclip -o -selection clipboard -t x-special/gnome-copied-files | grep --text --color=never file://"
```
stevalkr: @stevearc, the output of the command `xclip -o -selection clipboard -t x-special/gnome-copied-files` is similar to this:

```
copy
file:///path/to/file
```

To address this I added

```
grep —text —color=never file://"
```

to ignore the initial line "copy\n" (which could also be "cut" or something else). However, this change is not present in your commit, thus it's currently broken on my Ubuntu desktop.

stevearc: Should be fixed now.

stevalkr:
It’s working on all my machines now, thanks a bunch for your help!

stevearc:
Great! I'm going to try to make it work with yanking/pasting multiple files before merging. Might skip multi-file yank on Mac for now since it's such a PITA, but it looks like it should be straightforward on linux.

stevearc:
@stevalkr alright, could you give this one last test to make sure it's all still working as expected? Code paths to hit:

- Copy one file
- Copy multiple files
- Paste one file
- Paste multiple files
- Copy directory
- Paste directory

adelarsq:
Cool feature! I mainly use Windows  for development (no WSL). I will take a look about to support this on it too.

stevalkr:
@bew, I’m not certain if the built-in clipboard on Linux supports setting the MIME type beyond plain text. For macOS, pbcopy and pbpaste are incapable of copying files. Also, the MIME type can vary across different desktops. If you’re aware of how to implement this using the built-in clipboard, I’ll be glad to help ;)

Kyren223:
@stevalkr 

This feature doesn't seem to work, for context:
Commit: 685cdb4ffa74473d75a1b97451f8654ceeab0f4a
NVIM v0.11.0
KDE plasma 6.3.4 wayland (so wl-clipboard)

While in an oil buffer, I tried the following:
y$ yy Vy on a single file, all of which just copy the name, which only makes it paste-able in the local buffer, and it doesn't copy it into the clipboard.
Pasting an already copied file (copied with ctrl+c, not thru oil) and then pasting it into a buffer just pastes: file:///home/kyren/Desktop/a trying to save the buffer with :w shows the following errror:
```
Error detected while processing BufWriteCmd Autocommands for "oil://*":
Error executing lua callback: ....local/share/nvim/lazy/oil.nvim/lua/oil/mutator/init.lua:74: assertion failed!
stack traceback:
        [C]: in function 'assert'
        ....local/share/nvim/lazy/oil.nvim/lua/oil/mutator/init.lua:74: in function 'add_action'
        ....local/share/nvim/lazy/oil.nvim/lua/oil/mutator/init.lua:123: in function 'create_actions_from_diffs'
        ....local/share/nvim/lazy/oil.nvim/lua/oil/mutator/init.lua:571: in function 'try_write_changes'
        ...e/kyren/.local/share/nvim/lazy/oil.nvim/lua/oil/init.lua:985: in function 'save'
        ...e/kyren/.local/share/nvim/lazy/oil.nvim/lua/oil/init.lua:1240: in function <...e/kyren/.local/share/nvim/lazy/oil.nvim/lua/oil/init.lua:1227>
```

Kyren223:
I have not, now it seems to work, thanks!

Although that's not how I imagined it working, is there a way to have the normal copy/pasting (thru y and p) use this functionality instead?
If not then I guess it'd just be very inconvenient but at least it's a possibility and I only occasionally use it.

stevalkr:
@Kyren223 Neovim uses registers, not a single system clipboard (and it’s not the default one), to manage copied and pasted text. You can use the + register (e.g. "+y for normal text) to interact with the system clipboard for copy and paste. If you use system clipboard only, you can map the actions to `yy` and `p`. But it’s not possible for the plugin to determine which one to use

stevalkr:
> you can map the actions to `yy` and `p`.

This is exactly what you want, copying part of an entry is meaningless anyway. The system clipboard is more expensive and less versatile (it’s incompatible with SSH, for instance). It’s sensible to keep it as an alternative rather than the default.

Kyren223:
I agree that having clipboard be the default will be weird.

Although (and that's a separate issue/discussion) maybe having a different way of identifying files that uses absolute paths will be better, it'd be another option to achieve this (without copying the file into the clipboard which is OS-dependent) and it'd most likely resolve the occasional errors I get when saving oil buffers after local copy/pasting operations

stevalkr:
> maybe having a different way of identifying files that uses absolute paths will be better

This decision should be made by the maintainer (who isn’t me). However, it aligns with the original plan for this PR, to support absolute paths. I intended to make it work for all adapters, but I couldn’t get it to work with SSH. I might attempt it again when I have some free time ;)

LetsBlend:
I am glad this feature was implemented, it's exactly what i needed. I was wondering if it was also possible to implement cut operation in addition to the copy mechanism?

stevalkr:
@LetsBlend If I understand correctly, you're suggesting a macOS-style "cut": a normal copy, followed by a special "paste-and-delete-source" action.

I think there's a clean way to implement it. When pasting from the system clipboard, we already load the parent directory of the source file. This means we can create a new action that performs a move instead of a copy.

If that's what you're looking for, please open a new issue to track it. I'd be happy to help with the implementation when I'm available.

LetsBlend:
@stevalkr Yes that is precisely what I was suggesting. Created the issue: #648

PR #560, opened 2025-01-17
fix: restore previously focused window when opening file in multiple-split layout with an open preview window

MomePP:
This PR restores the original window when opening file using the oil floating window. There is an implementation to restore the focused window but only works for the toggle action.

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -709,7 +709,7 @@ M.select = function(opts, callback)
     else
       -- Close floating window before opening a file
       if vim.w.is_oil_win then
-        vim.api.nvim_win_close(0, false)
+        M.close()
       end
     end
 


lua/oil/init.lua
```diff
@@ -709,7 +709,16 @@ M.select = function(opts, callback)
...
       -- Close floating window before opening a file
       if vim.w.is_oil_win then
+        -- Store original winid before close oil floating window
+        local original_winid = vim.w.oil_original_win
+
+        -- Close floating window
         vim.api.nvim_win_close(0, false)
+
+        -- Restore previous focused window
+        if original_winid and vim.api.nvim_win_is_valid(original_winid) then
```
BozeBro, line 719: Can you give an example where the new wnid != current window id?

I was playing around with your PR, after we close the current window, the win_id seems to be the same as the original winid.

MomePP: in my case, there is a usage where multiple splits are used.

for example, we have 3 splits : top-left, bottom-left and one right split.
```
|-------|-------|
|       |       |
|-------|       |
|       |       |
|-------|-------|
```

Before the PR, when we focusing on the right split and opening a file using `Oil --float`, it would always open in the top-left split.

After the PR, it will correctly open a file in the previously focused split which is the right split.

MomePP: normally, i use the nightly, but i have tested with the latest release as well. the issue still occurs.
```
NVIM v0.10.3
Build type: Release
LuaJIT 2.1.1736781742
```

MomePP: Ah, it was my oil config. if i mapped my keybind to call only `oil.toggle_float()`. it behaves as aspected, opening file in the right split.

however, in my case, i configured my keybind to open a preview window right after `toggle_float` as shown.
``` lua
function()
    oil.toggle_float()
    oil_utils.run_after_load(0, function()
         if oil.get_cursor_entry() then
             oil.open_preview()
         end
    end)
end
```

or, if i open the float window first and then press <C-p> to open the preview window, the issue still remains.

BozeBro: tough catch. Close out the PR if everything works

MomePP: but that mean if we want to open a file in the correct split, we can't use the preview window ?
or do we need to close the preview window first before opening a file ?

i think this PR still needed to handle these edge cases of usage.

MomePP:
further more, i found out that it needs to call `M.close()` instead of handling it separately in `M.select()` as it is redundant to close the floating window.

lua/oil/init.lua
BozeBro: To open in the correct split, I believe oil already opens the file in the last focused window pane

BozeBro:
> further more, i found out that it needs to call `M.close()` instead of handling it separately in `M.select()` as it is redundant to close the floating window.

Ig I don't see how M.close handles the edge that nvim_win_close doesn't handle.
also for M.close, it seems the purpose of it is to close the buffer, and then restore the contents of the buffer before oil was opened.

In this new case, we want to close the buffer, and open a new file in the original window which is different

MomePP:
@BozeBro 

Maybe I was wrong about the root cause of this issue. So, I created a new issue where we can discuss it further. Once we identify the proper solution, we can update this PR to fix the issue.

https://github.com/stevearc/oil.nvim/issues/566

In this issue, I was able to replicate the behavior using a bare minimal config. So, you could try reproducing the issue as well.

stevearc:
Appreciate filing the issue as well, was able to repro myself and better understand the situation. Fix looks good, thanks!

Issue #536, opened 2024-12-11
bug: Cursor position doesn't auto align on directory enter when ../ (parent directory) item is shown


johnpgr:
### Did you check the docs and existing issues?

- [X] I have read the docs
- [X] I have searched the existing issues

### Neovim version (nvim -v)

v0.10.2

### Operating system/version

Ubuntu 24.04.1 LTS

### Describe the bug

When first entering a directory in Oil, the cursor gets automatically positioned to the first file name start.
That doesn't happen if the hidden files are enabled, because the first file is now `../` (parent directory)

### How it behaves with hidden files disabled

https://github.com/user-attachments/assets/7aa67544-3365-4847-bfe2-709ed6f85bc1

### How it behaves with hidden files enabled

https://github.com/user-attachments/assets/018fd0c1-4414-4da5-b7f4-f694407a1c56

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. toggle hidden files
2. navigate between directories

### Expected Behavior

The cursor position on first directory enter should have a consistent behavior with hidden files enabled.

### Directory structure

any directory structure

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"--single-branch",
		"https://github.com/folke/lazy.nvim.git",
		lazypath,
	})
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
	"folke/tokyonight.nvim",
	{
		"stevearc/oil.nvim",
		config = function()
			require("oil").setup({
				columns = {
					"permissions",
					"size",
					"ctime",
					"icon",
				},
			})
		end,
	},
	-- add any other plugins here
}
require("lazy").setup(plugins, {
	root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Did you check the bug with a clean config?

[...]

PR #562, opened 2025-01-20
fix: set cursor on .. directory when hidden files on and entering new directory

BozeBro:
PR Issue #536
ASSUMPTION: the id of `..` is `0`


stevearc:
Thanks for the PR! I've opted to solve it a different way, by re-using the existing `constrain_cursor` function (see [c80fa5c](https://github.com/stevearc/oil.nvim/commit/c80fa5c415b882c1c694a32748cea09b7dafc2c5))

Issue #522, opened 2024-11-28
feature request: send a subset of files to the quickfixlist


Janshagen:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

The ability to send all files in a directory to the quickfixlist was recently added (https://github.com/stevearc/oil.nvim/issues/223). An obvious extension would be to only send some of the selected files, either only the visually selected files, or something more clever

### Provide background

Example: You would like to do some quickfix-edit to some files in a directory, but there are also a lot of files that you do not want to edit. I want a good way to add only some of the files

### What is the significance of this feature?

nice to have

### Additional details

I have adapted the code mentioned above to work for visually selected lines (without breaking the prior behavior):

```lua
local function selected_oil_files_to_quickfix()
	if vim.bo.filetype ~= "oil" then
		return
	end

	local start_line = 1
	local end_line = vim.fn.line("$")

	local mode = vim.fn.mode()
	if mode == "v" or mode == "V" or mode == "\22" then
		start_line = vim.fn.line("v")
		end_line = vim.fn.line(".")

		if start_line > end_line then
			start_line, end_line = end_line, start_line
		end
	end

	local oil = require("oil")
	local dir = oil.get_current_dir()

	local entries = {}
	for line_number = start_line, end_line do
		local entry = oil.get_entry_on_line(0, line_number)
		if entry and entry.type == "file" then
			table.insert(entries, { filename = dir .. entry.name })
		end
	end

	if #entries == 0 then
		return
	end

	vim.fn.setqflist(entries)
	return vim.cmd.copen()
end
```

Janshagen:
Here is another version (that breaks the prior behavior) which adds the files that match the latest search pattern:

```lua
local function searched_oil_files_to_quickfix()
	if vim.bo.filetype ~= "oil" then
		return
	end

	local pattern = vim.fn.getreg("/")
	if pattern == "" then
		return
	end

	local oil = require("oil")
	local dir = oil.get_current_dir()

	local entries = {}
	for line_number = 1, vim.fn.line("$") do
		local entry = oil.get_entry_on_line(0, line_number)
		if entry and entry.name:find(pattern) and entry.type == "file" then
			table.insert(entries, { filename = dir .. entry.name })
		end
	end

	if #entries == 0 then
		return
	end

	vim.fn.setqflist(entries)
	return vim.cmd.copen()
end
```

Janshagen:
[...]

PR #564, opened 2025-01-22
Extend quickfix features

Janshagen:
I have added the features from this issue: https://github.com/stevearc/oil.nvim/issues/522

By passing an additional option to the send_to_qflist action, one can now choose whether to only add files that matches the last search when search highlighting is enabled.  

In doing so I also found a bug with choosing whether to add or replace the entries in the quickfix list

### DIFF
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -504,10 +504,12 @@ M.send_to_qflist = {
     opts = vim.tbl_deep_extend("keep", opts or {}, {
       target = "qflist",
       action = "r",
+      only_matching_search = false,
     })
     util.send_to_quickfix({
       target = opts.target,
       action = opts.action,
+      only_matching_search = opts.only_matching_search,
     })
   end,
   parameters = {
@@ -518,6 +520,10 @@ M.send_to_qflist = {
     action = {
       type = '"r"|"a"',
       desc = "Replace or add to current quickfix list (see |setqflist-action|)",
+    },
+    only_matching_search = {
+      type = "boolean",
+      desc = "Whether to only add the files that matches the last search. This option only applies when search highlighting is active",
     },
   },
 }
--- a/lua/oil/util.lua
+++ b/lua/oil/util.lua
@@ -753,7 +753,7 @@ end
 
 ---Send files from the current oil directory to quickfix
 ---based on the provided options.
----@param opts {target?: "qflist"|"loclist", action?: "r"|"a"}
+---@param opts {target?: "qflist"|"loclist", action?: "r"|"a", only_matching_search?: boolean}
 M.send_to_quickfix = function(opts)
   if type(opts) ~= "table" then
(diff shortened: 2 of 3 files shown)


lua/oil/actions.lua
```diff
@@ -519,6 +521,10 @@ M.send_to_qflist = {
       type = '"r"|"a"',
       desc = "Replace or add to current quickfix list (see |setqflist-action|)",
     },
+    match_all = {
+      type = "boolean",
+      desc = "Whether to only add the files that matches the last search",
+    },
```
stevearc, line 527: In keeping with the "all items are the default", could we rename this to `only_matching_search`? The current name only describes the default behavior, it doesn't tell you anything about what the behavior would be if the default is _not_ used.

lua/oil/util.lua
```diff
@@ -817,6 +819,19 @@ M.get_visual_range = function()
...
+---@param entry oil.Entry
+---@return boolean
+M.is_matching = function(entry)
+  -- if search highlightig is not enabled, all files are considered to match
+  local search_highlighting_is_off = (vim.v.hlsearch == 0)
+  if search_highlighting_is_off then
+    return true
+  end
+  local pattern = vim.fn.getreg("/")
+  local position_of_match = entry.name:find(pattern)
```
stevearc, line 831: This is doing a raw string find, which should work for searches like `/foo`, but I think this will break down for regex searches like `/foo.*.txt`. You probably want to use `vim.fn.match` or similar.

stevearc (changes requested):

lua/oil/actions.lua
Janshagen: I understand what you mean. The con of your suggestion is that the check in util becomes: `if entry and entry.type == "file" and (not only_matching_search or M.is_matching(entry)) then` (since the default values switches from true to false), which seems a bit less clear. But it is your call

lua/oil/util.lua
Janshagen: You are right, I will change

lua/oil/actions.lua
stevearc: Or you can pull that out into a variable above the if-statement?
```lua
local take_entry = not only_matching_search or M.is_matching(entry)
if entry and entry.type == "file" and take_entry then
  ...
end
```
If you don't like that you could also make another intermediate var
```lua
local match_all = not only_matching_search
```

Janshagen: Another alternative is this: `if entry and M.should_add_to_qflist(entry, opts) then`. And moving all the checking into a separate function. But maybe that is too vague

stevearc: Given that this won't be used anywhere else and it's just a couple lines, I think it's better to not extract to a separate function

stevearc:
Great, looks good thanks for the PR!

PR #565, opened 2025-01-22
Add hint to disable lazy loading in lazy.nvim install

### LINKED ISSUE #539: bug: Oil doesn't execute when Lazy.nvim needs to install something

- [X] I have read the docs

### Neovim version (nvim -v)

v0.10.2

### Operating system/version

MacOS 15.1.1

### Describe the bug

When running `nvim .`, and Lazy.nvim needs to install / update a plugin, Oil fails to open.

`nvim-tree` suggests adding the following lines to `init.lua`, but that doesn't help here

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1
```

I have tried with no Oil configuration, and with 
```lua
-- ...
opts = {
  default_file_explorer = true
}
-- ...
```

### Steps To Reproduce

1. `nvim -u repro.lua .`
2. Wait for Lazy.nvim to bootstrap / download the oil plugin

### Expected Behavior

I expected Oil to open the given directory.

This works when no plugins need to be installed, but if even one plugin needs to install or be updated, Oil fails to open after Lazy.nvim is done. 

### Directory structure

repro.lua

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              default_file_explorer = true
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

justmejulian:
Closes #539
Disabling lazy loading is necessary to set Oil as the default explorer when Neovim is initially opened.
Not 100% sure about the wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ oil.nvim supports all the usual plugin managers
   opts = {},
   -- Optional dependencies
   dependencies = { { "echasnovski/mini.icons", opts = {} } },
-  -- dependencies = { "nvim-tree/nvim-web-devicons" }, -- use if prefer nvim-web-devicons
-  -- Disable lazy loading which can interfere with features like 'default_file_explorer' 
-  -- lazy = false,
+  -- dependencies = { "nvim-tree/nvim-web-devicons" }, -- use if you prefer nvim-web-devicons
+  -- Lazy loading is not recommended because it is very tricky to make it work correctly in all situations.
+  lazy = false,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


stevearc:
Sure, seems fine to call this out explicitly

PR #574, opened 2025-01-31
pass bufnr to the constrain_cursor

### LINKED ISSUE #573: bug: error when another buffer gets quickly opened

- [x] I have read the docs

### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

MacOS 15 (Sequoia)

### Describe the bug

When I open editor with `nvim .` and quickly open a new buffer, I get the following error:

```console
Could not find oil adapter for scheme '://'
[oil] could not find adapter for buffer '://'
```

I investigated with `debug.traceback()` and that took me to `constrain_cursor` which does not seem to take `bufnr` and uses `0` instead in a couple of places.

https://github.com/stevearc/oil.nvim/blob/add50252b5e9147c0a09d36480d418c7e2737472/lua/oil/view.lua#L268
https://github.com/stevearc/oil.nvim/blob/add50252b5e9147c0a09d36480d418c7e2737472/lua/oil/view.lua#L274

Adding `bufnr` to `constrain_cursor` and forwarding them seems to fix the issue for me (I'll try to propose a PR).

I have searched for the issues, and I did find one issue that looked similar to what I was having: https://github.com/stevearc/oil.nvim/issues/362, but that seems to happen in a different place.

In the `repro` script below, I have set `vim.schedule(vim.cmd.new)` to quickly open a buffer.
But in reality, I usually do `nvim .`, and then immediately press a keymap to open a `snacks.picker` to find files. That is where I get this error. So, `vim.schedule(vim.cmd.new)` is just my attempt to imitate opening another buffer. :)

Also tested on Neovim 0.11.0 which is what I am using at the moment. But I can also reproduce in v0.10.4.

minor (annoyance)

### Steps To Reproduce

1. `nvim -u repro.lua .`

### Expected Behavior

No errors are displayed.

### Directory structure

_No response_

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
vim.schedule(vim.cmd.new)
```

skshetry:
Fixes #573.

### DIFF
--- a/lua/oil/view.lua
+++ b/lua/oil/view.lua
@@ -264,6 +264,9 @@ local function constrain_cursor(bufnr, mode)
   if not mode then
     return
   end
+  if bufnr ~= vim.api.nvim_get_current_buf() then
+    return
+  end
   local parser = require("oil.mutator.parser")
 
   local adapter = util.get_adapter(bufnr)
--- a/lua/oil/view.lua
+++ b/lua/oil/view.lua
@@ -258,20 +258,21 @@ local function get_first_mutable_column_col(adapter, ranges)
 end
 
 ---Force cursor to be after hidden/immutable columns
+---@param bufnr integer
 ---@param mode false|"name"|"editable"
-local function constrain_cursor(mode)
+local function constrain_cursor(bufnr, mode)
   if not mode then
     return
   end
   local parser = require("oil.mutator.parser")
 
-  local adapter = util.get_adapter(0)
+  local adapter = util.get_adapter(bufnr)
   if not adapter then
     return
   end
 
   local cur = vim.api.nvim_win_get_cursor(0)
-  local line = vim.api.nvim_buf_get_lines(0, cur[1] - 1, cur[1], true)[1]
+  local line = vim.api.nvim_buf_get_lines(bufnr, cur[1] - 1, cur[1], true)[1]
   local column_defs = columns.get_supported_columns(adapter)
   local result = parser.parse_line(adapter, line, column_defs)
   if result and result.ranges then
@@ -406,7 +407,7 @@ M.initialize = function(bufnr)
     callback = function()
       -- For some reason the cursor bounces back to its original position,
       -- so we have to defer the call
-      vim.schedule_wrap(constrain_cursor)(config.constrain_cursor)
+      vim.schedule_wrap(constrain_cursor)(bufnr, config.constrain_cursor)
     end,
   })
   vim.api.nvim_create_autocmd({ "CursorMoved", "ModeChanged" }, {
@@ -419,7 +420,7 @@ M.initialize = function(bufnr)
         return
       end
 
-      constrain_cursor(config.constrain_cursor)
+      constrain_cursor(bufnr, config.constrain_cursor)
 
       if config.preview_win.update_on_cursor_moved then
         -- Debounce and update the preview window
@@ -690,7 +691,7 @@ local function render_buffer(bufnr, opts)
             end
           end
 
-          constrain_cursor("name")
+          constrain_cursor(bufnr, "name")
         end
       end
     end)


lua/oil/view.lua
```diff
@@ -258,20 +258,21 @@ local function get_first_mutable_column_col(adapter, ranges)
...
   end
   local parser = require("oil.mutator.parser")
 
-  local adapter = util.get_adapter(0)
+  local adapter = util.get_adapter(bufnr)
   if not adapter then
     return
   end
 
   local cur = vim.api.nvim_win_get_cursor(0)
```
stevearc, line 274: If bufnr is not the current buffer, then 0 is the wrong winid to be using. Could we either pass the winid as well, or check to see if the current buffer == bufnr and early return?

Incidentally, do you know where exactly the problem is arising? The only time I see this getting executed async is the `vim.schedule` call inside of `InsertEnter`, but the original bug report didn't mention anything about insert mode. Is the problem that something is switching the buffer right after entering insert mode? Or that something is causing you to enter insert mode as you switch buffers?

stevearc (changes requested):
1 comment & Q

lua/oil/view.lua
skshetry: > If bufnr is not the current buffer, then 0 is the wrong winid to be using. Could we either pass the winid as well, or check to see if the current buffer == bufnr and early return?

Sure, I'll try.

> Incidentally, do you know where exactly the problem is arising? The only time I see this getting executed async is the `vim.schedule` call inside of `InsertEnter`, but the original bug report didn't mention anything about insert mode. Is the problem that something is switching the buffer right after entering insert mode? Or that something is causing you to enter insert mode as you switch buffers?

All the `vim.schedule` makes it hard to get a full stack trace. I added `print(debug.traceback())` in a bunch of places to get the following results (note the line number may be off by a couple of lines).


```lua
.../oil.nvim/lua/oil/init.lua:1221: in function <.../oil.nvim/lua/oil/init.lua:1220>
stack traceback:
        .../oil/view.lua:536: in function 'initialize'
        .../oil.nvim/lua/oil/init.lua:1079: in function <.../oil.nvim/lua/oil/init.lua:1057>
        [C]: in function 'nvim_win_call'
        .../oil.nvim/lua/oil/init.lua:1057: in function 'callback'
        .../oil.nvim/lua/oil/adapters/files.lua:265: in function ''
stack traceback:
        .../oil.nvim/lua/oil/view.lua:680: in function 'render_buffer'
        .../oil.nvim/lua/oil/view.lua:905: in function ''
stack traceback:
        .../oil.nvim/lua/oil/config.lua:467: in function 'get_adapter_by_scheme'
        .../oil.nvim/lua/oil/util.lua:66: in function 'get_adapter'
        .../oil.nvim/lua/oil/view.lua:269: in function 'constrain_cursor'
        .../oil.nvim/lua/oil/view.lua:700: in function <.../oil.nvim/lua/oil/view.lua:681>
Could not find oil adapter for scheme '://'
stack traceback:
        .../oil.nvim/lua/oil/util.lua:68: in function 'get_adapter'
        .../oil.nvim/lua/oil/view.lua:269: in function 'constrain_cursor'
        .../oil.nvim/lua/oil/view.lua:700: in function <.../oil.nvim/lua/oil/view.lua:681>
[oil] could not find adapter for buffer '://'
```

It starts from `setup()`, with a callback on `BufReadCmd`.
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/init.lua#L1220
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/init.lua#L1078
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/view.lua#L531
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/view.lua#L898
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/view.lua#L693
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/view.lua#L268
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/util.lua#L66
https://github.com/stevearc/oil.nvim/blob/abbfbd0dbcaa78c3dcdada191ea23e50a41e5806/lua/oil/config.lua#L468

skshetry: I don't switch to insert mode, but the snacks.picker does on it's buffer.

Also, here's a `repro.lua` that I can reproduce the bug from #573.

<details><summary>`repro.lua`</summary>
<p>

```lua 
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/oil.nvim",
    config = function()
      require("oil").setup({
      })
    end,
  },
  { "folke/snacks.nvim", opts = { picker = {} } },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- It takes a few try to reproduce with `vim.schedule`
vim.schedule(Snacks.picker.files)
-- Can be reproduced with following reliably
-- Snacks.picker.files()
```

</p>
</details>

stevearc: Got it, yeah it seems like the quick switch to the snacks picker and insert mode is likely the culprit here

Olical: I also reproduce this error when opening telescope quickly after oil loads. Just in case that helps at all.

The act of passing the bufnr and winnr across the schedule boundary makes total sense to me though, the ID of 0 should only ever be used in fully sync code.

skshetry: >If bufnr is not the current buffer, then 0 is the wrong winid to be using. Could we either pass the winid as well, or check to see if the current buffer == bufnr and early return?

I went with early return instead of passing `winid`, as `constrain_cursor` is synchronous.

Hernan-Ibarra: > I also reproduce this error when opening telescope quickly after oil loads. Just in case that helps at all.

I'm having the same issue (annoying because I have scripts that open directories and start telescope). Thanks to everyone who is working on fixing this.

stevearc:
LGTM thanks for the PR!

PR #575, opened 2025-02-01
Add support for bufnr in column rendering functions

Landaman:
This is primarily for user-defined custom columns, which may want access to the current path or similar information. The change is relatively simple since bufnr is necessarily always available when rendering columns, and will not break any consumers


stevearc (changes requested):
Seems fine to me! Looks like there was one legit CI failure, the rest will go away if you rebase latest master

Landaman:
Should be all set now, thanks!

stevearc:
Thanks for the PR!

PR #592, opened 2025-03-06
fix(trash-win): don't hang when `shellslash` is enabled

### LINKED ISSUE #590: bug: Trash yields forever on Windows with shellslash option set

- [x] I have read the docs

### Neovim version (nvim -v)

NVIM v0.10.4 Build type: Release LuaJIT 2.1.1713484068

### Operating system/version

Windows 11

### Describe the bug

Trying to trash something gets stuck on 1/x forever. Trying to go to the recycle bin gets stuck on "Loading" forever.

### Steps To Reproduce

Try to delete any file or direction after setting `shellslash`

### Expected Behavior

The file/directory to be moved to the trash.

### Directory structure

_No response_

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              delete_to_trash = true,
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

vim.opt.shellslash = true
```

TheLeoP:
Fixes #590

For some reason beyond my understanding, at least one of the following
things requires `noshellslash` to avoid the embeded powershell process to
send only `""` to the stdout (never calling the callback because
`"===DONE(True)==="` is never sent to stdout)

- `vim.fn.jobstart`
- `cmd.exe`
- `powershell.exe`

`shellslash` has been removed from Neovim nightly so I didn't put that much time into investigating why or how this is happening.

### DIFF
--- a/lua/oil/adapters/trash/windows/powershell-connection.lua
+++ b/lua/oil/adapters/trash/windows/powershell-connection.lua
@@ -28,6 +28,16 @@ end
 
 ---@param init_command? string
 function PowershellConnection:_init(init_command)
+  -- For some reason beyond my understanding, at least one of the following
+  -- things requires `noshellslash` to avoid the embeded powershell process to
+  -- send only "" to the stdout (never calling the callback because
+  -- "===DONE(True)===" is never sent to stdout)
+  -- * vim.fn.jobstart
+  -- * cmd.exe
+  -- * powershell.exe
+  local saved_shellslash = vim.o.shellslash
+  vim.o.shellslash = false
+
   -- 65001 is the UTF-8 codepage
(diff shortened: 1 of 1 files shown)


stevearc:
Given the silence from OP I'm going to assume this resolves their problem. Thanks for the fix!

PR #601, opened 2025-03-20
fix: indexing nil when env vars does not exist

Bekaboo:
When environment variable `XDG_SESSION_DESKTOP` does not exist, copy files to clipboard will throw a nil value error. This PR adds nil checks for these environment variables.


stevearc:
Thanks for the fix!

PR #610, opened 2025-04-01
feat: add file/directory owner and group column

### LINKED ISSUE #436: feature request: columns for owner and group

### Describe the feature

Title should be self explanatory: it would be nice to have columns for owner and group. An extra bonus would be to make the owner/group editable to  `chown` entries.

### Provide background

_No response_

nice to have

### Additional details

_No response_

### LINKED ISSUE #599: feature request: add user:group showing and manipulation

### Describe the feature

You can show and edit file permissions, but you can't show and change user:group of a file. It would be really helpful if such feature will be implemented.

### Provide background

_No response_

strongly desired

### Additional details

_No response_

tocariimaa:
For now it can only display it. I will keep working on it to support modifying the owner and group. For this we parse the `/etc/passwd` and `/etc/group` files, so it is UNIX only, although it also may not work on some unix-like systems, like Android on Termux, for instance so I wonder how could I deal with that.

closes #436, closes #599

### DIFF
--- a/lua/oil/adapters/files.lua
+++ b/lua/oil/adapters/files.lua
@@ -58,6 +58,9 @@ file_columns.ownership = {
   render = function (entry, _, _)
     local meta = entry[FIELD_META]
     local stat = meta and meta.stat
+    if not stat then
+      return columns.EMPTY
+    end
     local user = passwd.passwd_entries[stat.uid]
     local group = passwd.group_entries[stat.gid]
     if user ~= nil and group ~= nil then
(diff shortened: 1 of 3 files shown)


tocariimaa:
So parsing `/etc/passwd` (and `group`) will only work on Linux and BSD systems. For Windows and Mac not of course. A "proper solution" would be using the standard OS APIs (`getpwuid`, `getpwnam`) but of course we can't from Lua, not without C bindings of some kind...

PR #614, opened 2025-04-06
feat: Custom confirmation keymaps

alexandremassaro:
Add the possibility to set custom keycode to confirm changes


stevearc:
This adds more API surface, which I try to avoid unless it's really necessary. In this case, it's quite easy for users to customize the confirmation keymaps using an autocmd (see https://github.com/stevearc/oil.nvim/issues/114#issuecomment-1571332722). I appreciate the PR, but this is not a good fit.

PR #616, opened 2025-04-09
fix: using select tab will create extra a blank buffer.

RicardoKovd:
Problem
---
When using action.select.opts.tab to open a file, an extra blank buffer is created.

This blank buffer comes from the command `tabnew`, the command will create a blank buffer and empty window, but action.select.opts.tab just to using empty window, does not to using this blank buffer.

https://github.com/stevearc/oil.nvim/blob/302bbaceeafc690e6419e0c8296e804d60cb9446/lua/oil/init.lua#L743-L763

Reproduction
---
```bash
$ nvim --version
NVIM v0.11.0
Build type: None
LuaJIT 2.1.1736781742
Run "nvim -V1 -v" for more info
```
Typing `:Oil<CR><C-t>`
image

How to fix
---
`tabnew` has the same properties like `edit`, so you can just open a buffer directly like `edit #<bufnr>`.

```doc
:[count]tabnew [++opt] [+cmd] {file}
		Open a new tab page and edit {file}, like with |:edit|.
		For [count] see |:tabnew| above.
```
```doc
							*:edit_#* *:e#*
:e[dit] [++opt] [+cmd] #[count]
			Edit the [count]th buffer (as shown by |:files|).
			This command does the same as [count] CTRL-^.  But ":e
			#" doesn't work if the alternate buffer doesn't have a
			file name, while CTRL-^ still works then.
			Also see |++opt| and |+cmd|.
```
Typing `:Oil<CR><C-t>`
image

### DIFF
--- a/lua/oil/init.lua
+++ b/lua/oil/init.lua
@@ -749,21 +749,23 @@ M.select = function(opts, callback)
         vim.bo[filebufnr].buflisted = true
       end
 
-      local cmd = "buffer"
       if opts.tab then
-        vim.cmd.tabnew({ mods = mods })
-      elseif opts.split then
-        cmd = "sbuffer"
-      end
-      ---@diagnostic disable-next-line: param-type-mismatch
-      local ok, err = pcall(vim.cmd, {
-        cmd = cmd,
-        args = { filebufnr },
-        mods = mods,
-      })
-      -- Ignore swapfile errors
-      if not ok and err and not err:match("^Vim:E325:") then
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for the PR! This did have the side effect of _not_ wrapping the call to `tabnew` with pcall, so now I believe it is possible for opening a file in a new tab to generate swapfile errors. I took a simpler approach in https://github.com/stevearc/oil.nvim/commit/5b6068aad7d2057dd399fac73b7fb2cdf23ccd6e to clean up the buffer; hopefully it works for you!

Issue #617, opened 2025-04-13
feature request: filetype by actual filetype


nobe4:
### Did you check existing requests?

- [x] I have searched the existing issues

### Describe the feature

Enable Oil to find the filetype by nvim's computed filetype. Not only the file extension.

### Provide background

I use a lot of scripts in `./script/` for common functions (test, lint, serve, ...).
They all are without extension, so when I change repo/OS I know that `test` will always run the test, regardless of the underlying system.

This works well enough, but oil.nvim seems to be only using the filename extension to determine the filetype.

(n)vim can also use other markers, like the shebang line.
It would be awesome to integrate with `vim.filetype.match`'s `contents`. For example:

For instance:

```lua
:lua c = vim.fn.readfile("./script/test")
:lua f = vim.filetype.match({ filename = "test", contents = c })   
:lua vim.print(f) -- => sh
```

I have tried to do such changes in https://github.com/nobe4/oil.nvim/commit/64bf6523f8b220ce00f96660e5f545f60797f597
But it's probably incomplete.

### What is the significance of this feature?

nice to have

### Additional details

If this sounds like a worthy idea, I'd love to have some pointers towards making this a reality!

nobe4:
Edit, it seems that I misunderstood the function that I read earlier, the code that pulls the icon is actually: https://github.com/stevearc/oil.nvim/blob/302bbaceeafc690e6419e0c8296e804d60cb9446/lua/oil/util.lua#L928-L951

However, at no point does this seems to involve reading the file, and is relying on 3rd party instead.

stevearc:
This isn't explicitly stated anywhere in the documentation, but one thing that is very important to me about oil is the performance (see https://github.com/stevearc/benchmark.nvim/tree/master/examples/file_explorers). I work in large codebases that sometimes have _enormous_ directories and I want the load time to be as close to instant as possible. Reading the entire contents of each file in a directory before displaying it can potentially slow that down a _lot_.

I would potentially be okay with adding some logic around this, but only if it's carefully written to preserve responsiveness. For example, only running when there are <N entries in the directory, and only reading the first few lines of the file instead of the entire contents.

PR #618, opened 2025-04-13
feat(icons): filetype detection via shebang

nobe4:
This PR adds a step in the columns making: read the first line of the file and use that to compute a "fallback" filetype in case the filename alone cannot be used to determine the filetype.

This is then passed to the `IconProvider` which uses it if the icon is nil, or default.

This works well in initial tests, but `vim.fn.readfile(name, '', 1)` only read files in the current folder, I need to understand how I can make that work when changing folders. However, once this is done, I think this will provide some nice additional icons.

Looking forward to getting a review/suggestions :)

### DIFF
--- a/lua/oil/columns.lua
+++ b/lua/oil/columns.lua
@@ -168,7 +168,14 @@ if icon_provider then
       if meta and meta.display_name then
         name = meta.display_name
       end
-      local icon, hl = icon_provider(field_type, name, conf)
+
+      local ft = ""
+      if field_type == "file" then
+        local firstline = vim.fn.readfile(name, '', 1)
+        ft = vim.filetype.match({ filename = name, contents = firstline }) or ""
+      end
+
+      local icon, hl = icon_provider(field_type, name, ft, conf)
       if not conf or conf.add_padding ~= false then
         icon = icon .. " "
       end
--- a/lua/oil/columns.lua
+++ b/lua/oil/columns.lua
@@ -170,7 +170,7 @@ if icon_provider then
       end
 
       local ft = nil
-      if config.use_vim_ft then
+      if config.use_slow_filetype_detection then
         if field_type == "file" then
           local firstline = vim.fn.readfile(name, '', 1)
           ft = vim.filetype.match({ filename = name, contents = firstline })
--- a/lua/oil/columns.lua
+++ b/lua/oil/columns.lua
@@ -153,7 +153,7 @@ end
 local icon_provider = util.get_icon_provider()
 if icon_provider then
   M.register("icon", {
-    render = function(entry, conf)
+    render = function(entry, conf, bufnr)
       local field_type = entry[FIELD_TYPE]
       local name = entry[FIELD_NAME]
       local meta = entry[FIELD_META]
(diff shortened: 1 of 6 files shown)


lua/oil/util.lua
```diff
@@ -930,19 +930,26 @@ M.get_icon_provider = function()
   local _, mini_icons = pcall(require, "mini.icons")
   ---@diagnostic disable-next-line: undefined-field
   if _G.MiniIcons then -- `_G.MiniIcons` is a better check to see if the module is setup
-    return function(type, name)
-      return mini_icons.get(type == "directory" and "directory" or "file", name)
+    return function(type, name, ft)
+      local icon, hl, is_default = mini_icons.get(type == "directory" and "directory" or "file", name)
+      if icon == nil or is_default then
+        icon, hl = mini_icons.get("filetype", ft or "")
+      end
```
stevearc, line 937: Probably should only do this when `ft` is non-nil, right?

stevearc: Actually, I think for both devicons and MiniIcons it makes sense to use the filetype _first_ if it's provided, as that implementation should both be faster and more correct, and only use the original logic if the filetype is nil.

stevearc (changes requested):
I mentioned this [in the issue](https://github.com/stevearc/oil.nvim/issues/617#issuecomment-2927654702), but performance is a concern here. Passing a max lines of `1` to `readfile` is helpful, but this will still slow us down in large directories because we'll have to open a _lot_ of files. Could we maybe put this behind a config option so the default behavior is still fast?

lua/oil/columns.lua
```diff
@@ -168,7 +168,14 @@ if icon_provider then
       if meta and meta.display_name then
         name = meta.display_name
       end
-      local icon, hl = icon_provider(field_type, name, conf)
+
+      local ft = ""
+      if field_type == "file" then
+        local firstline = vim.fn.readfile(name, '', 1)
```
nobe4, line 174: @stevearc, I have searched, and failed to find how to get absolute filepaths, or the current directory in the context of this function. Do you know how I can get it?

lua/oil/util.lua
nobe4: I have pushed some commits to address this, mini.icons changed, but devicons actually was simplified, since it handles the `ft == nil` logic already.

lua/oil/columns.lua
stevearc: The render function takes a bufnr as the third argument
https://github.com/stevearc/oil.nvim/blob/685cdb4ffa74473d75a1b97451f8654ceeab0f4a/lua/oil/columns.lua#L15
You could parse the buffer name to get the parent dir like we do here
https://github.com/stevearc/oil.nvim/blob/685cdb4ffa74473d75a1b97451f8654ceeab0f4a/lua/oil/adapters/files.lua#L506-L507

lua/oil/config.lua
```diff
@@ -100,6 +100,9 @@ local default_config = {
       { "type", "asc" },
       { "name", "asc" },
     },
+    -- Read the first line of the file and use vim's matching ft.
+    -- Can slow down the execution considerably for large projects.
+    use_vim_ft = false,
```
stevearc, line 105: Could we instead make this an option for the `icon` column, like we do for `add_padding`? You can just access it on the `conf` table like `conf.use_slow_filetype_detection` and then add an entry to [this file](https://github.com/stevearc/oil.nvim/blob/685cdb4ffa74473d75a1b97451f8654ceeab0f4a/scripts/generate.py#L140-L152) to generate the documentation for it.

lua/oil/columns.lua
nobe4: Awesome! Thanks for this!

Done in https://github.com/stevearc/oil.nvim/pull/618/commits/e2ad451f57c5cfa82ef7bded0b44f473e98a2919

Issue #630, opened 2025-05-27
bug: When modifying and saving oil buffer with LSP integration, neovim throws E5108 (Invalid glob)


kmoschcau:
### Did you check the docs and existing issues?

- [x] I have read the docs
- [x] I have searched the existing issues

### Neovim version (nvim -v)

v0.12.0-dev-446+gc4e52d604c

### Operating system/version

Windows 11

### Describe the bug

I get this stack trace:

```
E5108: Lua: C:/Program Files/Neovim/share/nvim/runtime/lua/vim/glob.lua:373: Invalid glob
stack traceback:
	[C]: in function 'assert'
	C:/Program Files/Neovim/share/nvim/runtime/lua/vim/glob.lua:373: in function 'to_lpeg'
	.../Local/nvim-data/lazy/oil.nvim/lua/oil/lsp/workspace.lua:90: in function 'get_matching_paths'
	.../Local/nvim-data/lazy/oil.nvim/lua/oil/lsp/workspace.lua:282: in function 'will_rename_files'
	...ta/Local/nvim-data/lazy/oil.nvim/lua/oil/lsp/helpers.lua:84: in function 'will_perform_file_operations'
	...a/Local/nvim-data/lazy/oil.nvim/lua/oil/mutator/init.lua:396: in function 'process_actions'
	...a/Local/nvim-data/lazy/oil.nvim/lua/oil/mutator/init.lua:579: in function 'cb'
	...nvim-data/lazy/oil.nvim/lua/oil/mutator/confirmation.lua:135: in function 'confirm'
	...nvim-data/lazy/oil.nvim/lua/oil/mutator/confirmation.lua:187: in function <...nvim-data/lazy/oil.nvim/lua/oil/mutator/confirmation.lua:186>
```

Afterwards the whole oil buffer is no longer writable, but can still be used for file navigation.

I have to stop writing the repro steps here, because I am at work right now. I will have a look later, when I have more time, should the current ones not be enough.

### What is the severity of this bug?

breaking (some functionality is broken)

### Steps To Reproduce

1. have `ts_ls` installed
2. add a file `foo.ts` with `export default class Foo {}`
3. add a second file `boo.ts` with `import Foo from "./foo.ts"
4. Open an oil buffer
5. try to rename `foo.ts` to `bar.ts` and save
6. the error occurs

### Expected Behavior

There should be no error thrown.

### Directory structure

tsconfig.json
foo.ts
bar.ts

### Repro

```Lua
[...]

PR #631, opened 2025-06-04
fix: makes workaround conditional as it is no longer needed for 0.12

kaerumm:
Fixes https://github.com/stevearc/oil.nvim/issues/630 that happened because of https://github.com/neovim/neovim/commit/322a6d305d088420b23071c227af07b7c1beb41a


stevearc (changes requested):
LGTM! Please fix the CI failures

stevearc:
Thanks!

Issue #634, opened 2025-06-13
bug: parse ls line failed


jiz4oh:
### Did you check the docs and existing issues?

- [x] I have read the docs
- [x] I have searched the existing issues

### Neovim version (nvim -v)

NVIM v0.11.1

### Operating system/version

macOS 15.3.1

### Describe the bug

I am trying to connect my android phone by adb ssh by `nvim oil-ssh://civi//sdcard/Android` and an error raised as below:

```
|| ...files/vim/bundle/oil.nvim/lua/oil/adapters/ssh/sshfs.lua:172: table index is nil
|| stack traceback:
|| 	...files/vim/bundle/oil.nvim/lua/oil/adapters/ssh/sshfs.lua:172: in function 'cb'
|| 	.../vim/bundle/oil.nvim/lua/oil/adapters/ssh/connection.lua:232: in function '_handle_output'
|| 	.../vim/bundle/oil.nvim/lua/oil/adapters/ssh/connection.lua:131: in function <.../vim/bundle/oil.nvim/lua/oil/adapters/ssh/connection.lua:127>
```

the core reason is [the `line` parameter to the `parse_ls_line`](https://github.com/stevearc/oil.nvim/blob/08c2bce8b00fd780fb7999dbffdf7cd174e896fb/lua/oil/adapters/ssh/sshfs.lua#L162) has a difference datetime format like `"-rw-rw----   1 10197 1023   360 2025-04-05 14:27 .fd000021243cbfdbf4ef30c830952de5"`.

`"360 2025-04-05 14:27 .fd000021243cbfdbf4ef30c830952de5"` match failed with [`"^(%d+)%s+(%S+%s+%d+%s+%d%d:?%d%d)%s+(.*)"`](https://github.com/stevearc/oil.nvim/blob/08c2bce8b00fd780fb7999dbffdf7cd174e896fb/lua/oil/adapters/ssh/sshfs.lua#L48)

### What is the severity of this bug?

breaking (some functionality is broken)

### Steps To Reproduce

1. pass `"-rw-rw----   1 10197 1023   360 2025-04-05 14:27 .fd000021243cbfdbf4ef30c830952de5"` to the [`parse_ls_line`](https://github.com/stevearc/oil.nvim/blob/08c2bce8b00fd780fb7999dbffdf7cd174e896fb/lua/oil/adapters/ssh/sshfs.lua#L28)
2. no `name` has been returned

### Expected Behavior

please add compatible with the ios8601 format

[...]

PR #635, opened 2025-06-13
fix: add iso8601 format compatibility

jiz4oh:
fixes https://github.com/stevearc/oil.nvim/issues/634, fallback to match the `iso8601` format


stevearc:
Thanks for the fix!

PR #640, opened 2025-07-01
doc: add a mention to third party extension oil-git.nvim

benomahony:
Adding a mention of another 3rd party extension for colouring the filenames in the oil buffer based on git status
Screenshot 2025-07-01 at 09 59 20


stevearc:
LGTM thanks!

PR #652, opened 2025-08-03
fix: support natural ordering for numbers with >12 digits

XeroOl:
Changes the column ordering code when `view_options.natural_order` is enabled, so that it can support larger numbers.

I chose to handle numbers of up to 999 digits, because some cursory research seems to suggest to me that most filesystems cap at 255 bytes per segment, and 999 > 255, so it should be sufficient for all actual filenames.

Performance may have changed but I don't know if it will be better or worse.

I've slightly increased the complexity of the `string.format` format specifier, but the final strings will be shorter on average, and `string.format` no longer has to implicitly coerce `int` from a string into a number.

before:
image
after:
image

### DIFF
--- a/lua/oil/columns.lua
+++ b/lua/oil/columns.lua
@@ -228,8 +228,8 @@ M.register("type", {
   end,
 })
 
-local function pad_number(int)
-  return string.format("%012d", int)
+local function adjust_number(int)
+  return string.format("%03d%s", #int, int)
 end
 
 M.register("name", {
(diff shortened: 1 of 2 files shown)


XeroOl:
With the original natural order PR you said you did benchmarking. I'm curious what the methodology was, because I want to know if I've improved performance at all

stevearc:
Nice work! I appreciate the iteration on performance

PR #662, opened 2025-08-26
feat: virtual column headers

cosmicbuffalo:
This adds a couple new optional settings `show_header` and `header_format`. the first is a boolean that turns on a header row displayed with virtual text, and the second is a function that takes each header string and should return a new string for a header display. if the `header_format` function is not set, the upcased column name is used by default.

an basic example config:
```lua
{
  show_header = true,
  header_format = function(header_text)
    if header_text == "MTIME" then
      return "Modified"
    elseif header_text == "CTIME" then
      return "Changed"
    elseif header_text == "ATIME" then
      return "Accessed"
    elseif header_text == "BIRTHTIME" then
      return "Created"
    elseif header_text == "SIZE" then
      return "Size"
    elseif header_text == "NAME" then
      return "Filename"
    else
      return header_text
    end
  end
  -- other options...
}
```

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -51,6 +51,8 @@ local default_config = {
   constrain_cursor = "editable",
   -- Set to true to watch the filesystem for changes and reload oil
   watch_for_changes = false,
+  -- Show column headers at the top of the oil buffer using virtual text
+  show_header = false,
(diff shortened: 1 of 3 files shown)


cosmicbuffalo:
I think this approach is still half baked, I'll need to think about it some more before I open again

PR #677, opened 2025-10-27
S3 support via a new adapter

### LINKED ISSUE #633: feature request: Adapter for accessing s3 buckets

### Describe the feature

It would be very nice to be able to use oil for looking through s3 buckets similar to how I can use it for looking through files on another machine with ssh.
 
I don't know how feasible this is, but I think it should be doable with `aws s3 ls` commands and similar. (Of course this would then require a proper setup of `aws`, but that can be expected from anyone wanting to use this feature.)

To start with I think the main feature for s3 would be just file navigation and file inspection (without editing). I think it should be possible to add editing too, but that is less important. 

I don't know if you would consider adding this to oil itself or if it fits better in a third party extension. If you are open to adding it, I might be able to look into making an initial implementation (though I don't know how much time I will have for this).

### Provide background

I prefer navigating all files I look at with oil, and this is the only type of files I can't navigate currently.

nice to have

### Additional details

I did try https://github.com/kiran94/s3edit.nvim and it works fine for smaller s3 buckets, but the commands used take way too long on some larger s3 buckets (which is why I recommend `aws s3 ls` since I checked that it runs faster), and it doesn't allow me to navigate with oil as I am used to (in particular I can't easily go in and out of files).

Danielkonge:
This is still a draft and not completely ready yet, but I might need a bit of help with some of the non-S3 related things (see below). [Update: It is mostly ready now and I think it makes sense to get some feedback before I do much more on it.]

s3 is a bit different from ssh since we don't keep a constant connection and instead make new `aws s3 ls` calls when moving around, but I have still tried to build similar behavior to what is currently in `sshfs.lua`, see `s3fs.lua`.

Having the "filesystem" setup similarly, we then also build the adapter similar to what is done for ssh, see `ssh.lua` and `s3.lua`. (Again only with minor differences from the fact that we don't keep an active connection in the same way.)

I am still running into a few problems, among which the primary of the top of my head is:
- I had to temporarily use the scheme `oil-sthree://` since any scheme with a number in its name made it so that neovim didn't open the buffer correctly. I am not sure how to best work around this, but `oil-s3://` would be the clearly better scheme name here.
- The mutator is a bit hard to work with from the adapters perspective. It would probably be nicer to move some of its logic to each individual adapter instead.
- There are still a few minor bugs.
- I think there should be a better way to start the Oil s3 view from an already open neovim session, maybe just with `Oil --s3` or similar, but I haven't added that yet.
- Not really a problem, but more of a note: I haven't updated the README and docs yet. And I have a few `vim.notify` calls to track behavior that should of course be removed before this is ready.

For testing this out I made a new aws account and used `extra_s3_args = { "--profile=<my-profile>", "--region=<my-region>" }` when running stuff. 

If you want to edit some of this yourself if that is easier, feel free to edit directly on this branch. (Also let me know if you want cleaner commits, then I can just rebase.)

This PR will close #633.

### DIFF
--- a/lua/oil/adapters/s3.lua
+++ b/lua/oil/adapters/s3.lua
@@ -164,25 +164,11 @@ M.normalize_url = function(url, callback)
   callback(url_to_str(res))
 end
 
-M.has_awscli_checked = false
----@return boolean
-M.has_awscli = function()
-  if M.has_awscli_checked then
-    return true
-  end
-
-  if vim.fn.executable("aws") == 1 then
-    M.has_awscli_checked = true
-    return true
-  end
-  return false
-end
-
 ---@param url string
 ---@param column_defs string[]
 ---@param callback fun(err?: string, entries?: oil.InternalEntry[], fetch_more?: fun())
 M.list = function(url, column_defs, callback)
-  if not M.has_awscli() then
+  if vim.fn.executable("aws") ~= 1 then
     callback("`aws` is not executable. Can you run `aws s3 ls`?")
     return
   end


lua/oil/adapters/s3.lua
```diff
@@ -0,0 +1,343 @@
...
+  -- default assumption is that everything is modifiable
+  return true
+end
+
+---@param action oil.Action
+---@return string
+M.render_action = function(action)
+  if action.type == "create" then
+    local extra = action.type == "bucket" and "BUCKET " or ""
+    return string.format("CREATE %s%s", extra, action.url)
```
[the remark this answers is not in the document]
Danielkonge: This was just because of a typo `type` instead of `entry_type`, it works fine now.

lua/oil/init.lua
```diff
@@ -1109,6 +1121,13 @@ local function close_preview_window_if_not_in_oil()
   pcall(vim.api.nvim_win_close, preview_win_id, true)
 end
 
+---@param pat string Pattern to escape
+---@return string Escaped pattern
+local function escape_pattern(pat)
+  local res = pat:gsub("-", "%%-")
+  return res
+end
```
Danielkonge: Reading other filetype patterns and inspecting `vim.filetype` a bit, I think we need to escape `-` in `oil-ssh` and `oil-s3`. (Just for the filetype patterns, not for the autocmd patterns.

stevearc:
If the region is only really needed for creating new buckets, then maybe we should just not allow the creation of buckets using oil. It seems like a pretty uncommon use case; I would imagine that most of the time this would be used for exploring and manipulating buckets that already exist.

If we also give up the special icon/highlight for buckets then I don't think we would need the new entry type at all. I know it's a slight downgrade in usability, but it keeps all the functionality and significantly decreases the complexity of this change. We might come up with some other way to customize the icons/highlights in the future, since this is something that people keep asking for.

lua/oil/init.lua
Danielkonge: Should I just remove this from this PR? It is unrelated to the s3 support.

Danielkonge:
> If the region is only really needed for creating new buckets, then maybe we should just not allow the creation of buckets using oil. It seems like a pretty uncommon use case; I would imagine that most of the time this would be used for exploring and manipulating buckets that already exist.

I think I might have confused a bit with what I wrote earlier. The region is generally not needed, and you can create buckets without the region. The only time you might run into problems is when you manually switch aws account with the `--profile` flag, but this is already an uncommon use case mostly for more advanced users, and they should then also be able to figure out how to add the region correctly. At most I think we might want to mention this in the docs, but I am not even sure that is worth it for such an edge case.

> If we also give up the special icon/highlight for buckets then I don't think we would need the new entry type at all. I know it's a slight downgrade in usability, but it keeps all the functionality and significantly decreases the complexity of this change. We might come up with some other way to customize the icons/highlights in the future, since this is something that people keep asking for.

I have removed the entry type and icon/highlight for buckets now, and everything should still work fine. 

I also updated the README and docs now.

lua/oil/config.lua
```diff
@@ -204,9 +206,14 @@ local default_config = {
 -- The adapter API hasn't really stabilized yet. We're not ready to advertise or encourage people to
 -- write their own adapters, and so there's no real reason to edit these config options. For that
 -- reason, I'm taking them out of the section above so they won't show up in the autogen docs.
+
+-- not "oil-s3://" on older neovim versions, since it doesn't open buffers correctly with a number
+-- in the name
+local oil_s3_string = vim.fn.has("nvim-0.12") and "oil-s3://" or "oil-sss://"
```
stevearc: `has()` returns a 1 for true and 0 for false, which are both truthy in lua. You'll want to do `vim.fn.has("nvim-0.12") == 1 and ...`

Issue #669, opened 2025-09-29
bug: `actions.select` 's `close` option does not work


superatomic:
### Did you check the docs and existing issues?

- [x] I have read the docs
- [x] I have searched the existing issues

### Neovim version (nvim -v)

NVIM v0.11.4

### Operating system/version

Fedora Linux 42

### Describe the bug

When selecting a file with a keybind using the `close = true` option, the buffer will not close.

### What is the severity of this bug?

breaking (some functionality is broken)

### Steps To Reproduce

1. Open a directory with `nvim -u repro.lua .`
2. Use `<CR>` to select a file or directory.
3. Run `:buffers`. The original oil buffer will still be open.

### Expected Behavior

According to the `oil-actions` documentation, I expect this option to `Close the original oil buffer once selection is made`.

### Directory structure

a/b

*(Any two-level file hierarchy will reproduce the bug consistently)*

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "stevearc/oil.nvim",
    config = function()
      require("oil").setup({
        buf_options = {
          buflisted = true,
          bufhidden = '',
        },
        cleanup_delay_ms = false,
        keymaps = {
          ['<CR>'] = {
            'actions.select',
            opts = {
              close = true,
            },
          },
        },
      })
    end,
  },
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
```

### Did you check the bug with a clean config?

- [x] I have confirmed that the bug reproduces with `nvim -u repro.lua` using the repro.lua file above.

PR #680, opened 2025-10-31
Fix oil.select close handling when reusing current window

dev-KingMaster:
## Summary
- ensure `oil.select` closes all remaining Oil windows when `close = true`
- delete the Oil buffer after closing if no windows remain, preventing stale buffers
- add a regression test to cover the buffer deletion behavior when reusing the current window

Issue link: https://github.com/stevearc/oil.nvim/issues/669
## Testing
- ./run_tests.sh tests/select_spec.lua


stevearc:
The `close` parameter is currently working as intended. The verbiage is a little tricky but, "close" in this case does not mean "delete", it means calling `oil.close()` which will restore the previous buffer to the window and background the oil buffer. Oil already has some logic that will periodically clean up unmodified oil buffers if they are not being displayed.

If you're looking for a way to explicitly delete oil buffers, I _have_ been meaning to make a change and just haven't found the time for it. I want to add a method `delete_all_buffers()` that will explicitly trigger the cleanup process. It'll be a bit more involved than the current process because it will need to properly handle oil buffers that are still displayed in windows, but it should be relatively straightforward. If you want to take a crack at that, I would happily review a PR

PR #681, opened 2025-11-03
Feat: configurable buffer cleanup for deletes/moves (closes #207)

### LINKED ISSUE #207: feature request: suppress "no longer available"

### Describe the feature

When deleting a file that is open, the corresponding buffer is still loaded and you get a "no longer available" message.
It would be nice to have an option that would automatically delete buffers that correspond to deleted files, and save+reopen buffers for the moved files.

### Provide background

Moving/deleting open files is annoying atm.

### Additional details

Got the idea from `tpope/vim-eunuch`s `:Remove`

dev-KingMaster:
## Title
feat: configurable buffer cleanup for deletes/moves (closes #207)

## Summary

- Adds optional buffer cleanup that:
  - Closes unmodified buffers when their files (or directories) are deleted.
  - Reloads buffers when their files are moved.
  - Can force-close modified buffers when desired.

## Config

- New buffer_cleanup option in require('oil').setup() (disabled by default):
- Example:
```
buffer_cleanup = {
delete = true, -- close buffers for deleted files/dirs
move = true, -- reload buffers for moved files
force = false, -- apply to modified buffers too
}
```

## Changes

- Implement cleanup logic in lua/oil/buffer_cleanup.lua and invoke it after actions.
- Wire into mutation flow via lua/oil/cache.lua.
- Documented in README.md and doc/oil.txt.
- Added tests in tests/buffer_cleanup_spec.lua.

## Notes

- Default behavior remains unchanged unless buffer_cleanup is enabled.
- Handles Windows path normalization and directory recursive deletes.

### DIFF
--- a/tests/select_spec.lua
+++ b/tests/select_spec.lua
@@ -91,13 +91,5 @@ a.describe("oil select", function()
       assert.equals(bufnr, vim.api.nvim_get_current_buf())
     end)
 
-    a.it("deletes oil buffer when reusing window", function()
-      test_util.oil_open()
-      local oil_bufnr = vim.api.nvim_get_current_buf()
-      -- Go to the bottom, so the cursor is not on a directory
-      vim.cmd.normal({ args = { "G" } })
-      a.wrap(oil.select, 2)({ close = true })
-      assert.is_false(vim.api.nvim_buf_is_valid(oil_bufnr))
-    end)
   end)
 end)


lua/oil/buffer_cleanup.lua
```diff
@@ -0,0 +1,175 @@
+local config = require("oil.config")
+local fs = require("oil.fs")
+local util = require("oil.util")
+
+local M = {}
+
+---@class oil.BufferCleanupOptions
+---@field delete boolean
+---@field move boolean
+---@field force boolean
```
stevearc: This should be defined in the config.lua with a corresponding version where all the fields are optional to match what `setup()` expects, and the descriptions for what each of these does should be on each field.

lua/oil/buffer_cleanup.lua
```diff
@@ -0,0 +1,175 @@
...
+  local opts = {
+    delete = not not cleanup.delete,
+    move = not not cleanup.move,
+    force = not not cleanup.force,
+  }
+  if not opts.delete and not opts.move then
+    return nil
+  end
+  return opts
+end
```
stevearc: just why

PR #687, opened 2025-11-13
fix `send_to_quickfix` opening qf when the target is `loclist`

znxuz:
Just a 3-lines change to open the correct list based on the given target.


lua/oil/util.lua
```diff
@@ -823,11 +823,12 @@ M.send_to_quickfix = function(opts)
   local action = opts.action == "a" and "a" or "r"
   if opts.target == "loclist" then
     vim.fn.setloclist(0, {}, action, { title = qf_title, items = qf_entries })
+		vim.cmd.lopen()
   else
     vim.fn.setqflist({}, action, { title = qf_title, items = qf_entries })
+		vim.cmd.copen()
```
bew: ⚠️ you're using tabs instead of spaces like the rest of the code here

PR #691, opened 2025-11-25
fix: `:belowright hor 10Oil`

phanen:
* Support more command mods `:belowright hor 10Oil`
* Fix `:tab Oil` only work on the first tab

Extract inline function to make the diff small after format.


phanen:
@looshch fixed in https://github.com/stevearc/oil.nvim/pull/695.

PR #694, opened 2025-12-06
fix!: use `g~` over `~`

seblyng:
I am opening a PR to see if you are open to this. This will essentially be a breaking change, but I have seen a couple of issues around this, and also that you have regretted the decision to use `~` as a mapping.

~Specifically for me, it was kind of annoying that `<C-l>` and `<C-h>` was mapped by default since I have those mapped to `<C-w>l` and `<C-w>h`~

~which I also would say are quite regular default mappings~

Let me know what you think about this! Hope you can consider!


seblyng:
Hmm okay, the maparg thing didn't seem to work as I wanted, so I will revert that. Hope you will consider the `~` mapping though

Issue #672, opened 2025-10-04
bug: Invalid glob: **/*.{}


Bekaboo:
### Did you check the docs and existing issues?

- [x] I have read the docs
- [x] I have searched the existing issues

### Neovim version (nvim -v)

v0.12.0-dev-1326+g35fc4fda99

### Operating system/version

Linux

### Describe the bug

When LSP is attached, sometimes I get the following error when deleting/moving/copying a file in oil buffers:

<img width="1073" height="333" alt="Image" src="https://github.com/user-attachments/assets/f48e0a93-4003-4eeb-bddf-afad03a658e8" />

I haven't found a way to reliably reproduce this, but this happens to me >1 time every day. The fix should be simple by checking if `glob` is empty in

https://github.com/stevearc/oil.nvim/blob/919e155fdf38e9148cdb5304faaaf53c20d703ea/lua/oil/lsp/workspace.lua#L91

### What is the severity of this bug?

minor (annoyance)

### Steps To Reproduce

1. Enter oil buffer in nvim with LSP attached
2. Try to remove/add a file in oil buffer, you can sometimes see the error.

### Expected Behavior

Should not error

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
              -- add any needed settings here
            })
        end,
  },
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

### Did you check the bug with a clean config?

[...]

PR #702, opened 2025-12-27
fix: handle empty LSP glob patterns

djdv:
fixes: #672
I have a [screencast](https://www.youtube.com/watch?v=ETdbe828e6M) of the issue + this patch.

[The LSP spec for glob patterns](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#fileOperationPattern) defines the syntax for "*group sub patterns*" but doesn't mention anything about empty patterns.

Somehow (not investigated) we end up with an empty `s` parameter being passed to our `glob` function, so I assume that since such a pattern isn't explicitly forbidden, that language servers (TSserver, marksman, et al.) emit them?
(Maybe this is worth checking up front if `s` is empty, but `if #filtered == 0 then` catches this. Might be even better to not call this function at all in some higher scopes but I'm not sure.)

Imagine patterns such as: `**/*.{md,}`, `**/*.{}`, `**/*.{foo,,bar}`.
These patterns get rejected by the parser for `vim.glob.to_lpeg()`: `Invalid glob: **/*.{}`.

---

This change replaces `glob` so that patterns are always sanitized and conditionally sorted before being passed to `to_lpeg()`.
So that `*.{}` is treated like `*`, `*.{md,}` -> `*.{md}`, `*.{foo,,bar}` -> `*.{foo,bar}`.
I believe these are all permissible in the current spec, so we account for them.

---

There may be a better way to solve this, I'm not sure.
This works for me in my use case (`NVIM v0.12.0-dev-1908+g47aef025a7` on Windows) but could use some testing from others.
Older vim users, people reporting this issue on other platforms + LSP servers and scenarios, etc.
Hopefully this approach doesn't break any existing workflows.

### DIFF
--- a/lua/oil/lsp/workspace.lua
+++ b/lua/oil/lsp/workspace.lua
@@ -79,9 +79,9 @@ local function get_matching_paths(client, filters, paths)
         glob = glob:gsub("{(.-)}", function(s)
           local patterns = vim.split(s, ",")
           local filtered = {}
-          for _, pattern in ipairs(patterns) do
-            if pattern ~= "" then
-              table.insert(filtered, pattern)
+          for _, pat in ipairs(patterns) do
+            if pat ~= "" then
+              table.insert(filtered, pat)
             end
           end
           if #filtered == 0 then


lua/oil/lsp/workspace.lua
```diff
@@ -76,17 +76,26 @@ local function get_matching_paths(client, filters, paths)
       ---@type string|vim.lpeg.Pattern
       local glob_to_match = glob
       if vim.glob and vim.glob.to_lpeg then
-        -- HACK around https://github.com/neovim/neovim/issues/28931
-        -- find alternations and sort them by length to try to match the longest first
-        if vim.fn.has("nvim-0.11") == 0 then
-          glob = glob:gsub("{(.*)}", function(s)
-            local pieces = vim.split(s, ",")
-            table.sort(pieces, function(a, b)
+        glob = glob:gsub("{(.*)}", function(s)
```
stevearc: We should replace the pattern with `{(.-)}` so that it's non-greedy. This is a bug with my original implementation.

PR #711, opened 2026-01-09
feat: add support for column text alignment

malewicz1337:
hey,

i was customizing my setup and realized I really wanted to right-align the file sizes and permissions—it just makes the table look cleaner, so I added a simple `align` option to the column config

it supports `left`, `center`, and `right`, so the default is still left, and it doesn't change existing behavior

i've updated the parsing logic to handle the extra whitespace and added it to the docs

p.s.
if you feel this adds unnecessary complexity or isn't something you want in the core, no hard feelings at all—feel free to just close this pr

thanks for the awesome plugin)))

### DIFF
--- a/doc/oil.txt
+++ b/doc/oil.txt
@@ -432,7 +432,7 @@ icon                                                                 *column-ico
                      the icon
 
 size                                                                 *column-size*
-    Adapters: files, ssh
+    Adapters: files, ssh, s3
     Sortable: this column can be used in view_props.sort
     The size of the file
 
@@ -485,7 +485,7 @@ atime                                                               *column-atim
       {format}    `string` Format string (see :help strftime)
 
 birthtime                                                       *column-birthtime*
-    Adapters: files
+    Adapters: files, s3
     Sortable: this column can be used in view_props.sort
     The time the file was created
 
@@ -600,6 +600,12 @@ preview                                                          *actions.previe
 
 preview_scroll_down                                  *actions.preview_scroll_down*
     Scroll down in the preview window
+
+preview_scroll_left                                  *actions.preview_scroll_left*
+    Scroll left in the preview window
+
+preview_scroll_right                                *actions.preview_scroll_right*
+    Scroll right in the preview window
 
 preview_scroll_up                                      *actions.preview_scroll_up*
     Scroll up in the preview window
--- a/lua/oil/actions.lua
+++ b/lua/oil/actions.lua
@@ -136,6 +136,30 @@ M.preview_scroll_up = {
   end,
 }
 
+M.preview_scroll_left = {
+  desc = "Scroll left in the preview window",
+  callback = function()
+    local winid = util.get_preview_win()
+    if winid then
+      vim.api.nvim_win_call(winid, function()
+        vim.cmd.normal({ "zH", bang = true })
+      end)
+    end
+  end,
+}
+
+M.preview_scroll_right = {
+  desc = "Scroll right in the preview window",
+  callback = function()
+    local winid = util.get_preview_win()
+    if winid then
+      vim.api.nvim_win_call(winid, function()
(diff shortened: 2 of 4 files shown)


lua/oil/util.lua
```diff
@@ -122,6 +122,32 @@ M.lpad = function(text, length)
   end
 end
 
+---@param text string
+---@param width integer|nil
```
stevearc: Is there any case where we need to call this with a `nil`?

lua/oil/util.lua
```diff
@@ -122,6 +122,32 @@ M.lpad = function(text, length)
...
+---@param align oil.ColumnAlign
+---@return string padded_text
+---@return integer left_padding
+M.pad_align = function(text, width, align)
```
stevearc: Let's also delete the old `util.rpad` function and replace it fully with this

lua/oil/util.lua
```diff
@@ -314,11 +340,14 @@ M.split_config = function(name_or_config)
...
 end
 
+---@alias oil.ColumnAlign "left"|"center"|"right"
 ---@param lines oil.TextChunk[][]
 ---@param col_width integer[]
+---@param col_align? oil.ColumnAlign[]
 ---@return string[]
 ---@return any[][] List of highlights {group, lnum, col_start, col_end}
-M.render_table = function(lines, col_width)
+M.render_table = function(lines, col_width, col_align)
```
stevearc: I think it makes more sense to only have one singular alignment value for a table that affects all columns, unless you have a case where you want some columns to be left aligned and some to be right aligned?
That would also mean that the config option should move from the columns onto the `view_options`

stevearc (changes requested):
This is a pretty minor feature, but it's a pretty small implementation as well. I'm happy to add something like this given that it doesn't add much code and shouldn't impact performance.

lua/oil/util.lua
```diff
@@ -344,12 +377,12 @@ M.render_table = function(lines, col_width)
...
               #str_lines,
-              col + sub_hl[2],
-              col + sub_hl[3],
+              col + padding + sub_hl[2],
+              col + padding + sub_hl[3],
             })
           end
         else
-          table.insert(highlights, { hl, #str_lines, col, col_end })
+          table.insert(highlights, { hl, #str_lines, col + padding, col + padding + text_len })
```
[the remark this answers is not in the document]
malewicz1337: i also looked at this several times and it concerned me too, but i didnt find a better way to do it without complicating it more. 

it was `col`, `col_end` where `col_end` = `col` + `text_len` + 1. the +1 was there cuz `col_end` was calculated after text had padding, so `text:len()` included padding, and now `text_len` is the original length before padding

i can probably add a comment explaining the calculations or rename variables?

lua/oil/util.lua
malewicz1337: yes, i thought different columns could be aligned differently, for example, size column -> right-aligned, permissions -> left-aligned, name -> left-aligned

lua/oil/util.lua
malewicz1337: yes, the test in `tests/parser_spec.lua` 89-100, passes `{}` to util.render_table() fn

lua/oil/columns.lua
```diff
@@ -97,6 +97,8 @@ end
 ---@return nil|string
 M.parse_col = function(adapter, line, col_def)
   local name, conf = util.split_config(col_def)
+  local leading_ws = line:match("^(%s*)")
+  line = line:sub(#leading_ws + 1)
```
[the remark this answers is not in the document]
malewicz1337: [5ae6ca7](https://github.com/stevearc/oil.nvim/pull/711/commits/5ae6ca7c9332d5df561e06537dfa895a234281ff)

malewicz1337:
my pleasure, happy to contribute))

PR #712, opened 2026-01-10
fix: propagate errors in recursive_delete and recursive_copy

SOberhoff:
The `complete` callback in `recursive_delete` and `recursive_copy` checks `err` instead of `err2`, but `err` is always nil inside the `elseif entries` branch. This silently ignores child operation errors, causing misleading "directory not empty" failures.

The correct pattern is already used elsewhere in the codebase:
- `recursive_move` (line 362)
- `trash/windows.lua` (line 336)


stevearc:
Good find, thanks for the fix!

Issue #304, opened 2024-02-05
feature request: Add option to specify preview split direction


omerchn:
### Did you check existing requests?

- [X] I have searched the existing issues

### Describe the feature

A `preview.split_direction` option to specify in which direction the preview split (`<C-p>`) opens

### Provide background

When I opened the preview split with `<C-p>` it always used to open to the right, but now after updating Oil it always opens to the left 🤔 
I would really appreciate the option to specify the direction. Thank you!

### What is the significance of this feature?

nice to have

omerchn:
I also noticed now that the `<C-s>` functionality also flipped - the vertical split is created to the left instead of the right 🤔 
Is this expected?

omerchn:
closing this as I no longer use oil

dudicoco:
@stevearc can we reopen this please?
I'm also interested in this feature.

moyukhc:
+1 to re-opening this request, I'd love to be able to tile my preview windows from NE as opposed to NW splits. I think you could just expose the `nvim_open_win.anchor` [parameter](https://neovim.io/doc/user/api.html#nvim_open_win()) from the neovim API docs?

Many thanks!

Florian-Merle:
Setting `vim.opt.splitright = true` works. See https://github.com/stevearc/oil.nvim/blob/master/lua/oil/init.lua#L471

Ideally, it would be great to have an oil option for that though

AdrianZajkowski:
I have similar issue with the `<C-s>` functionality. It opens vertical split always on the left. Does anyone have solution how to fix it and open split on the right side?

PR #713, opened 2026-01-10
Enable preview-split for non-floating view

abeaufays:
Hello ! I was facing the same problem as described [here](https://github.com/stevearc/oil.nvim/issues/304), but figured an alternative solution  that doesn't make me modify Oil.

I still got to a point where I could easily select the direction for vertical splits, I could not figure out how to configure it so it would be horizontal so I can't confirm it works for that.

I don't know if it matches the standards of this repo or good practices of Lua, but as it could be a nice addition I thought I would still share what I have.
I have an alternative solution so feel free to close the PR !

### DIFF
--- a/lua/oil/config.lua
+++ b/lua/oil/config.lua
@@ -155,6 +155,8 @@ local default_config = {
     end,
     -- Window-local options to use for preview window buffers
     win_options = {},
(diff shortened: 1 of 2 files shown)


stevearc:
I'd prefer to keep the current logic where you can pass the split options in directly. I think having two ways to customize the split behavior will lead to more confusion. Thanks for the thought though!

PR #716, opened 2026-01-17
Add `open_float` params to `toggle_float`

zeta-squared:
This is a straight-forward and small PR to add the param options available in `open_float` in `toggle_float`.


stevearc:
LGTM thanks!

PR #734, opened 2026-02-22
docs: combine setup and migration FAQ entries

barrettruth:
## Problem

The README FAQ had two separate entries for setup and migration that
covered overlapping ground.

## Solution

Merge them into a single entry with a before/after showing the
`opts`/`setup()` to `vim.g.oil`/`init` migration path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,28 +43,37 @@ luarocks install oil.nvim
 
 ## FAQ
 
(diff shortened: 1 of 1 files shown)


barrettruth:
@stevearc Horrible mistakes by me. Incorrectly set up my `gh` cli to point to the upstream. Really sorry for the bother. I've created a mechanism to hard-stop this behavior.

PR #739, opened 2026-03-07
feat(recipes): MacOS FreeDesktop trash configuration

ruuzia:
I looked into parsing` .DS_Store` files. It does not seem worth it. Hence, I thought this recipe would be useful. It is short, but not intuitive, and it works miraculously.


stevearc:
Hadn't thought people would want that, but it makes sense!

PR #744, opened 2026-04-01
fix: use os path when watching for changes

BrownBuilding:
Call `fs.posix_to_os_path` on the path parsed from the `oil://*`-url before handing it off to uv.fs_event_start. Not doing so breaks `watch_for_changes` on MS Windows.

(On Windows `oil://C/mydir/subdir/` would cause `uv.fs_event_start` to be called with `C/mydir/subdir/` rather than `C:\mydir\subdir\`)


BrownBuilding:
the top level drive view breaks on windows

stevearc:
Thanks for the fix!

PR #748, opened 2026-04-10
fix: add bounds check in calc_constrained_cursor_pos to prevent index…

zhu-jiyuan:
… out of bounds

   When buffer content changes during async re-rendering (e.g. loading
   animation replacing directory listing), CursorMoved/ModeChanged autocmds
   can fire before the cursor position is adjusted to the new buffer line
   count. This causes nvim_buf_get_lines with strict_indexing=true to throw
   "Index out of bounds" at view.lua:267.

   Add a line count check before accessing buffer lines so that
   calc_constrained_cursor_pos returns nil when the cursor is beyond the
   buffer, letting Neovim clamp the cursor on its own.
   
   
image

### DIFF
--- a/lua/oil/view.lua
+++ b/lua/oil/view.lua
@@ -264,6 +264,10 @@ end
 --- @return integer[] | nil
 local function calc_constrained_cursor_pos(bufnr, adapter, mode, cur)
   local parser = require("oil.mutator.parser")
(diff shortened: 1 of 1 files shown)


stevearc:
Thanks for the fix!

PR #756, opened 2026-05-27
doc: add installation instructions via vim.pack

emiara:
As of version 0.12, Neovim now has a built in plugin manager. Simply added it to the list. 👍🏽


stevearc:
LGTM thanks!

PR #779, opened 2026-09-03
fix: crash on case-only filename renames (case-insensitive filesystems)

### LINKED ISSUE #778: bug: invalid buffer error on renaming

- [x] I have read the docs

### Neovim version (nvim -v)

0.12.5

### Operating system/version

MacOS 26.6.2

### Describe the bug

i get this error on rename (the rename is successful though)

```
vim.schedule callback: .../anurag/.local/share/nvim/lazy/oil.nvim/lua/oil/util.lua:212: Invalid buffer id: 6                      
stack traceback:                                                                                                                  
        [C]: in function '__index'                                                                                                
        .../anurag/.local/share/nvim/lazy/oil.nvim/lua/oil/util.lua:212: in function <.../anurag/.local/share/nvim/lazy/oil.nvim/l
ua/oil/util.lua:194>
```

minor (annoyance)

### Steps To Reproduce

here's how it happens for me: 
suppose im in directory `~/projects/x`. 
1. i make a file `test.java`. i open it and write some code.
2. i create a new tab in my (kitty) terminal and run `javac test.java` which gives error `file should be named Example` or something like that. 
3. i switch back to this tab with neovim, open oil buffer, edit `test.java` to `Test.java` and `:w`. 

### Expected Behavior

No error

### Directory structure

_No response_

### Repro

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
        "stevearc/oil.nvim",
        config = function()
            require("oil").setup({
            })
        end,
  },
  -- add any other plugins here
}
  root = root .. "/plugins",
})

-- add anything else here
```

theAnuragMishra:
Renaming a file via oil where only the case changes (`test.java` ->
`Test.java`) throws `invalid buffer` error on a case-insensitive filesystem, but only if the file was already
open in a buffer.

Cause:
`vim.fn.bufadd(dest_buf_name)` resolves to the *same* buffer as
`src_bufnr` when the two names differ only in case, since they point
to the same inode. `rename_buffer`'s swap logic assumes `src_bufnr`
and `dest_bufnr` are always distinct, so it deletes `src_bufnr` and
then accesses `dest_bufnr` (the now-deleted same buffer) right after.

Fix:
Short-circuit when `dest_bufnr == src_bufnr`: just rename the buffer
in place and clean up the old-name buffer.

fixes #778


theAnuragMishra:
@stevearc can you check this out?

